PR #24, opened 2015-05-08
Update of index and download html files

smokingwheels:
My first request here goes.
index.hml
I added a link to the front end on the backends home page so that it is convenient to access it like how the webclient has a link to the back end.
Not sure if you want to use the port 3001 later on.

download.html
I changed the cd loklak in the setup instructions to loklak_server to reflect the repository. 
I added  instructions to download and run loklak server on a Debian Virtual Machine.
[Test Server](http://smokingwheels.mooo.com:9100)


sudheesh001:
This looks good to merge. @Orbiter I tested this PR. r+ @smokingwheels

smokingwheels:
I just removed the port 3001 on loklak.net since it is not used anymore.
My test server is off line for a while.

Issue #55, opened 2015-06-25
New Push API with support for other data formats


zyzo:
This Push API is different from current [push.json](https://github.com/loklak/loklak_server/blob/master/src/org/loklak/api/server/PushServlet.java) in several points : 
* it supports others formats than search result's one (geojson, ..)
* In POST parameters, users provide a link to their data file, not the data itself
* For geojson type, a `maptype` parameter must be specified to map attributes in geojson property fields. The maptype consists of simple map rules `geojson-field:loklak-tweet-field`, separated by commas.

Related : 
https://github.com/loklak/loklak_webclient/issues/205
https://github.com/loklak/loklak_webclient/issues/206
https://github.com/fossasia/api.fossasia.net/issues/45

Orbiter:
Please support this feature with some example input records, if available. I.e. how would a single geojson event from fossasia api look like?

zyzo:
@Orbiter This is the public link to fossasia api geojson file http://api.fossasia.net/map/ffGeoJsonp.php

Typically, the `properties` field will contain, if available, all informations enlisted here, except for location : 
https://github.com/loklak/loklak_webclient/issues/206#issuecomment-114064797

Orbiter:
interesting to see the `mtime` attribute there, because a reference time would be something which can be used to compose a unique ID for each message together with the location coordinates. This is needed to identify double-entries in subsequent submission of same records. We don't want to have several entries for the same event as messages, only one.

So I tend to say that giving a date in the property is obligatory, because otherwise a ID computation is not possible. Would you agree to that?

zyzo:
Yeah it's a good idea. The `mtime` is copied from `lastchange` attribute in our original json data. Maybe @andi will have more inputs on this. 

And when the file is updated (`mtime` changes), will the old file be discarded / overridden ? I guess that isn't the case for twitter tweets.

Orbiter:
we don't want to overwrite messages for the same location with a new date to be able to browse through time when visualizing this on a map. The idea is, that the map gets also a time-slider. Therefore the time is essential for the identification of the tweet.

On the other hand, I would like to give uploaders the opportunity to fix their mapping after an upload by just uploading again with a new mapping. In case that only the mapping has changed but not the records, they must be recorgnized with an ID which is independent from the given mapping.

andibraeu:
```lastchange``` is an attribute a community sets to their file, when they change anything in there
[...]

PR #56, opened 2015-06-26
New Push API

zyzo:
Add a geojson push endpoint to `/api/geojsonpush.json`.

There are several implementation details I want to discuss about this update : 

* The message `id_str` is the sum of longitude\*1000, latitude\*1000 and `mtime` in Unix epoch time. The *1000 exists to increase precision, since longitude and latitude are double values and must be converted to integer. `mtime` property is required, otherwise an message error with error code 400 with be sent.

* `text` property is currently not required, even though without it, creating a `MessageEntry` will give `NullPointerException`. What I suggests is make it mandatory like `mtime`.

* `source_type` property should be specified as POST parameter by user, if not, it will be equal to `IMPORT`. The implication is that it could exist in the db `source_type` values not listed in `org.loklak.harvester.SourceType`.

It would be helpful to have your opinion on these matters.

### DIFF
--- a/src/org/loklak/LoklakServer.java
+++ b/src/org/loklak/LoklakServer.java
@@ -47,6 +47,7 @@ import org.loklak.api.server.HelloServlet;
 import org.loklak.api.server.PeersServlet;
 import org.loklak.api.server.ProxyServlet;
 import org.loklak.api.server.PushServlet;
+import org.loklak.api.server.GeoJsonPushServlet;
 import org.loklak.api.server.SearchServlet;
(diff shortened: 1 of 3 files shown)


zyzo:
Related to https://github.com/loklak/loklak_server/issues/55

Orbiter:
I just merged your code and changed it a bit. I changed the api path to have a schema for more push apis for other formats, it's now at `/api/push/geojson.json`

Orbiter:
Sorry, made a mass when missing the latest commits. Now everything should be included.

Issue #58, opened 2015-06-27
Custom source_type values 


zyzo:
When creating a `MessageEntry` with value not listed in `SourceType` enum, the value of `source_type` is set to `USER`. 
What if I want to add other source types ? Is it mandatory that it has to be added to `SourceType` enum? Another way to do this is allow custom values to be added and searched.

zyzo:
The current strategy I'm adopting is to add new fields into `SourceType` enum class. That way, loklak codebase is in control of what source_type value is allowed and what not. Cf. branch [source-type-support](https://github.com/loklak/loklak_server/tree/source-type-support).

Orbiter:
The current source_type values are sufficient for the geojson task. The source is either a person who types in data in a front-end (USER), or it is provided by TWITTER or it was IMPORTed. As far as I can see, the source type for objects from geojson imports should be IMPORT.

zyzo:
I guessed I envisioned source_type value to be used for another purpose : distinguish sources of imported data. How do you do this (distinguish sources) ? This is important to retrieve messages from a single source, and display them on a map for example.

Orbiter:
oh I suggested this to Mario, yes it should be possible to identify the source. One piece of information that a geojson does not provide is the user name. My suggestion is to use the domain from the import as artificial author. For example: if the import comes from a url of the domain freifunk.de, then the artificial user name could be set as $freifunk_de. Then all such sources can be retrieved with the search term `from:$freifunk_de`. The `$` would then denote all such aritficial user names.

mariobehling:
@Orbiter: I would rather prefer to follow the approach to "attach" data sources to "real" user accounts. 

There are two ways to handle data imports in regards to user accounts:
1. We handle this like a general purpose search engine and add data from the web -- For me this is a possible future milestone
2. In a user generated service all data is "attached" to a user. -- This is the way I see the first version of loklak.

In this case the same data source can be added by different users into their account and they can do all kinds of things with it and build custom maps. This is the same if you compare that to how users and images work. The same image can be added in different tweets and uploaded by different users independently into their account.

A related question now is: How do we get a lot of data into loklak? This can be done by automating account creation and adding relevant data sources (semi)automatically. When there is useful data on loklak, we will get more people joining the service.

Does this make sense to you?

PR #59, opened 2015-06-29
source_type support

zyzo:
Support for source_type in both import & seach operations, specifically : 
  * allow importers to specify source_type value with parameter `source_type` (push/geojson.json)
  * add `source_type` search filter : `q=.../source_type=FOSSASIA_API`


zyzo:
Example of query to extract all fossasia api information : 

    http://localhost:9000/api/search.json?q=/source_type=FOSSASIA_API&count=200&minified=true

zyzo:
source_type will be from now on used to identify sources https://github.com/loklak/loklak_server/issues/58#issuecomment-120537076. Merging.

PR #65, opened 2015-07-04
Initial fixes towards replace existing words which have wrongly encoded characters

sudheesh001:
Works to replace bad text in single file at the moment


smokingwheels:
Do you have a processed file I can check and split with QB 4.5?
I used to have to delete some messages right at the beginning of the dumps.

sudheesh001:
@smokingwheels This patch is old, I might have to check it up and i'll report back.

Orbiter:
I consider this outdated

Issue #74, opened 2015-07-09
Implement FOSSASIA API JSON validator to loklak


zyzo:
After discussion with @mariobehling and @andibraeu, we think it would be best to have a json validator & converter, when pushing API messages to loklak. Right now, the user needs to provide a geojson data source. This raises several problems : 

* not every source is geojson.
* there's no constraint on the message format.
* most importantly, user is confused about what format they should use. Currently we have a [generator](http://api.fossasia.net/generator/index.html) that generates valid api json format, and our codebase is built around this format. If user provides valid api json, that makes it easy to import data from loklak -> fossasia. 

This doesn't apply solely to FOSSASIA API data, but other sources as well. The new workflow could look like this : 

* user pushes message, specifying that the message is for FOSSASIA API
* loklak_server calls corresponding validator. If the format is correct, call converter to convert into MessageEntry
* save the message into elastic

This model is extensible. I could build a validator layer that takes two files : a json specification ([example](https://github.com/fossasia/api.fossasia.net/blob/master/specs/1.0.1.json)), and a mapping file to map json field <-> MessageEntry field. Then, for each service that we want to provide closer support (like FOSSASIA), we just need to write these 2 files. 

@Orbiter what do you think about this proposal ?

zyzo:
More about JSON schema http://json-schema.org/

Orbiter:
We don not generate a *validator*. And of course not every source is geojson. We have to build a special import mechanism for each json schema that we want to import. When creating such an import mechanism, we look out for common formats where it is worth doing such an import process or we decide that the format is incompatible to your idea of having a message stream.

The FOSSASIA API schema describes events, therefore we can import them. But we need  special import process. Thats the reason I renamed your geojson api servlet into /api/push/geojson.json to have a concept where to add more import processes, like here it would be /api/push/fossasia.json.

We cannot have generic import processes, they must always consider the special background of the content of the json data. Therefore there is no use staring a discussion about validation etc. If we read json, it is alwys validated if the import process was able to import them. But that is not the point. The real point is the *evaluate* the data, not *validate* the data format.

zyzo:
>  If we read json, it is alwys validated if the import process was able to import them. But that is not the point.

[...]

Issue #35, opened 2015-06-04
provide push API for statical information


Orbiter:
To add more sources than harvested by twitter, we want to add data from other sources including RSS feeds and geoJSON data. These sources must be added to the message index in the context of a lifetime flag https://github.com/loklak/loklak_server/issues/33

The data submitted to the API must therefore include:
- URL of the source
- data format of the source (i.e. RSS/GeoRSS/geoJSON etc)
- a harvesting frequence (the submitter knows best how often the data is changed)
- a lifetime. The lifetime must be smaller or equal to the harvester frequence. The lifetime is asserted to the index and it may mean that the data disappears from search results after that time. A special lifetime of 2^31-1 can be set to announce that the data is statical forever, like a normal 'news' message, or a location that will never change (i.e. place of a city)

Orbiter:
create https://github.com/loklak/loklak_server/issues/37 first

zyzo:
Hi, what needs to be done to achieve this ? A new Push API is created (#55) but data is saved as MessageEntry without a harvesting frequence or a lifetime as described above.

Orbiter:
implemented with api/push/geojson.json

zyzo:
I have several blocker questions, that are critical to implement the connect service interface : 

* what field name is source url saved to ? same for harvester frequence ?
* Is the periodic harvester implemented yet ? So I just save the harvester frequence as a message field, and it will be automatically detected and harvested by the server ? 
* what if, one source url contains multiple messages ? Is it better to save common information (url, harvesting frequence) as new data type, rather than duplicate it in each message ? It would be much easier to update the source.

Orbiter:
> * what field name is source url saved to ? same for harvester frequence ?

nowhere yet. There must be a new data structure to hold this. At this time, data can just read from that url and if the import shall start again, the api must be called again. That if of course not the target design. It is true that the url must be stored and then the harvesting frequence must be either submitted as well or the frequence must be computed by try.

> * Is the periodic harvester implemented yet ? So I just save the harvester frequence as a message field, and it will be automatically detected and harvested by the server ? 

[...]

PR #77, opened 2015-07-13
Fossasia push service

zyzo:
Implementation of Fossasia API push service, via /api/push/fossasia.json.

The implementation follows 2-steps mechanism described in #74 : validate & convert fields. This makes integration of future services trivial : only 2 json files are needed per service : 
  
* a json schema
* and a mapping file to map original fields to tweet fields, so that the message can be properly saved into back-end). 

If you look at FossasiaPushServlet.java, it's mostly "high-level code" that deals with request and response. All json manipulating logic is handled by the validation & conversion layer.

TODO : 
- [x] In conversion step, allow deep mapping (`name` -> `user.screen_name`). This is done before with parameter map_type of /api/push/geojson.json
- [x] FOSSASIA API Json file don't have mtime field. Currently, the id is computed from only two values : lat & lon, which is not a secure strategy. A proper way to compute id is needed.
- [ ] link api data to rich content https://github.com/loklak/loklak_server/issues/55#issuecomment-118917334
- [ ] save statistical information #35 
- [x] allow a source to contains multiple documents (array type)


zyzo:
@Orbiter I have a question about the id computing strategy : is that true, that only ids from the same source are required to be different ? From an application point of view, messages from different sources, for e.g. TWITTER and FOSSASIA, with same ids cause no problem, since they can be distinguished by source_type. But I'm not sure how it works when saving to elastic db.

If this is true then I can just calculate the id based on lat & lon & community name. It is sure there's no two communities with the same name, located at the same area.

mariobehling:
@zyzo For FOSSASIA communities I do not see that we could have that case - same name same location, but for other data (people, devices) it might be possible that you get the same name in the same location. The question is, if you need to stay generic here or not.

Orbiter:
@zyzo 
> @Orbiter I have a question concerning id computing strategy : is that true, that only ids from the same source are required to be different?

no, all IDs must be unique, not only for a specific source

> From an application point of view, messages from different sources, for e.g. TWITTER and FOSSASIA, with same ids cause no problem, since they can be distinguished by source_type. But I'm not sure how it works when saving to elastic db.

we must make a source-prefix for the id, so they will be unique all the time.

> If this is true then I can just calculate the id based on lat & lon & community name. It is certain there's no two communities with the same name, located at the same area.

Lets consider the id as
`<source-domain>_<lon>_<lat>_<mtime>`
where mtime is the modification time that we must find out anyhow from the source. To prevent that we have a lot of doubles with same values, we must add a match there as well, but that is an add-on work.

zyzo:
> where mtime is the modification time that we must find out anyhow from the source.

For sources that already have `mtime`, this is easy. For those who don't, a solution is to register the first mtime as the time when the message is first added. When adding later version of the message, compare the content with latest version, if the content changes, save the message, with new mtime value.

We could save & compare content hashes to reduce computing cost, like git does.

The problem is can we find the latest version of a message ? A.K.A is it possible to find a message by its partial id  `<source-domain>_<lon>_<lat>_` ?

Orbiter:
> For sources that already have `mtime`, this is easy. For those who don't, a solution is to register the first mtime as the time when the message is first added. When adding later version of the message, compare the content with latest version, if the content changes, save the message, with new mtime value.

yes!

> We could save & compare content hashes to reduce computing cost, like git does.

If we compare the whole text part, then that would be the appropriate way to handle this, yes.

> The problem is can we find the latest version of a message ? A.K.A is it possible to find a message by its partial id  `<source-domain>_<lon>_<lat>_` ?

I planned to use the `<source-domain>` as screen_name with a special prefix, like `$<source-domain>`. Then it is easy to find the message by the screen_name and location_point.

zyzo:
> I planned to use the <source-domain> as screen_name with a special prefix, like $<source-domain>. Then it is easy to find the message by the screen_name and location_point.

Actually, we have source_type that also identifies source domain (for FOSSASIA it's fossasia_api). I made the source_type a constraint (PR #59), which means it is searchable by source_type like this : q=/FOSSASIA_API, or q=-/TWITTER.

So we can find the message by source_type and location_point ?

zyzo:
Also, for sources that already have id, should the format be `<source-domain>_<id>_<lon>_<lat>_<mtime>` ?

zyzo:
This is good to merge. Two remaining tasks will be implemented in the same time with other push services.

zyzo:
@Orbiter Could you help reviewing & merging this please ?

Orbiter:
I just looked through your code and found very interesting concepts and a lot of code. Thank you for this work! This looks like it nicely fits into the overall structure of the application, so it was a perfect pull request.

zyzo:
Thank you !

PR #83, opened 2015-07-23
Save import information (ImportProfile) when pushing custom messages

zyzo:
This PR introduces a new data type : ImportProfileEntry, that contains metadata information of a push : 

```java
protected String id;
protected Date created_at;
// importer username
protected String screen_name;
// importer ip address
protected String client_host;
protected URL source_url;
protected SourceType source_type;
protected int harvesting_freq;
protected int lifetime;
// id list of imported messages
protected List<String> imported;
```

Comparing to #35, there are some extra fields (screen_name, client_host, & created_at). This data is saved each time a push is successful, and contains at least 1 new message. `api/import.json` is used to retrieve the import profiles by `source_type`.

WHAT IS MISSING : 
- [x] a strategy to compute unique id. I wonder if there is a auto-generating id feature provided by elasticsearch.
- [x] a way to modify the import profile, once the profile is save. This feature could be embedded in /api/import.json, with a parameter e.g. `action=update`.

- information about data format of the source. As now we provide only `json` push, this feature can be safely postponed (until rss push, xml push.. are implemented).


zyzo:
It seems like when the id field is missing, it is auto-generated https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#_automatic_id_generation. 

But I think it's best to identify an import profile by its url and importer (screen_name) `<source_url>-<screen_name>`. The problem is, the user is not required to authenticate to twitter, to be able to push messages. This should be the default behaviour.

zyzo:
So the new id computing strategy is : 
 
* if screen_name is specified (via GET/POST parameter), the id is set to `<source_url>_<screen_name>`
* if not, it is set to `<source_url>_<client_host>`

I don't know if it is optimal (for e.g, when updating the screen_name of a profile without screen_name, things got complicated), but it ensures that an import profile is uniquely identified by its source url & its importer. @Orbiter @mariobehling what do you think ?

Orbiter:
>* if screen_name is specified (via GET/POST parameter), the id is set to `<source_url>_<screen_name>`
>* if not, it is set to `<source_url>_<client_host>`

> don't know if it is optimal (for e.g, when updating the screen_name of a profile without screen_name, things get complicated), but it ensures that an import profile is uniquely identified by its source url & its importer. 

I believe this is not sufficient, as IoT data may change over time while `<source_url>`, `<client_host>` and `<screen_name>` stays the same. That means we also need an id based on the content of the message. I suggest to compute a hash of the message content `<message_hash>` and add this to the ID as well

zyzo:
With appended message hash, a filter is added to search import profiles to display only latest result. Another concern is how user delete the import profile. I'm thinking a `status` field that will be set to `DELETED` to all versions of the same import profile.

zyzo:
- [ ] since #84 is merged saving import profile should be implemented in all push servlets.

zyzo:
@Orbiter could you please review & merge this ? Thanks.

zyzo:
(reposted from https://github.com/loklak/loklak_webclient/pull/394. Changes will be updated here as the branch is still not merged)

To implement the new import process, following things are needed from server:

* [ ] ability to let users specify `harvesting_freq` & `lifetime` at import time (when pushing things to /api/push/*.json)

* [ ] the list of source types & related information as an api endpoint. Currently the list is hardcoded as an [angular service](https://github.com/loklak/loklak_webclient/blob/data-connect/app/js/services/sourceType.js). 

* [ ] the list of loklak message field, to be able to implement an user-friendly field mapping step.

PR #110, opened 2015-08-10
Fix #99 saving message relies on mtime

### LINKED ISSUE #99: Save imported message only when message content changes
When `mtime` field is present, search for latest version of the message in elastic & compare `mtime` field. Save new message if `mtime` field differs. 

When the `mtime` field is not present, check the message existence in elastic. Compare content with old versions' & save new message only if content differs.

For safety reason (never rely on user) I think it's even better to apply strategy 2 to all cases, no matter if message contains `mtime` field or not.

Experimental branch : [id-with-mtime](https://github.com/loklak/loklak_server/tree/id-with-mtime)
See previous discussions : #77

zyzo:
During a new message import, check if message already exists in elastic db with no field differences. If it's the case, mark message as known and do not save. Otherwise save the message with new mtime value appended to message id `<source_type>_<lat>_<lon>_<mtime>`.

To be able to check message existence, the server performs search by source_type and location value. A new way to use location constraint is added - search by exact geo-point : 
```
q=/location=lat,lon
```

Fix #99

See also : https://github.com/loklak/loklak_server/pull/77#issuecomment-121219647


zyzo:
@Orbiter could you please review if this matches your thought on how we should handle mtime value as we discussed ? Thanks!

PR #113, opened 2015-08-16
Proposed fix for #112 message sharing feature

### LINKED ISSUE #112: Share message in Loklak
This feature is proposed by @mariobehling in https://github.com/loklak/loklak_webclient/issues/449. The idea is to implement a share feature so that people can keep track of a connection, watch change & tweet about change if they want. 

As of now when user add a connection, two things are saved : the messages, and an import profile containing metadata about the push. The import profile contains the list of message ids as well. My suggestion is to simply implement a "sharers" list in the import profile. @Orbiter what are your thoughts on this ?

zyzo:
Add `sharers` as an ImportProfileEntry field. Sharing feature is simply adding user screen_name to `sharers`  list.
Fix #112.


zyzo:
This is ready to merge.

PR #137, opened 2015-12-07
Added a python script for searching for tweets

pythad:


Orbiter:
Hey pythad, thank you!
Please add a `#!/usr/bin/env python` into the first line, which is a common way to trigger the right interpreter.

When I do that and test the script with `./search.py test` the response is
```
Traceback (most recent call last):
  File "./search.py", line 3, in <module>
    import requests
ImportError: No module named requests
```
This is not a bug but it would be nice to have a comment line which describes which dependencies in a fresh system must be installed and maybe how. Can you please add these details?

pythad:
Added requirements and  a `#!/usr/bin/env python` into the first line. I am curious if we need to use python2 instead of python3, does it makes sense?

Orbiter:
very well, that comment is exactly the right thing for python noobs to get this running. The script works as expected. Good work!

PR #148, opened 2015-12-10
Add a query profiling script

sevazhidkov:


sevazhidkov:
That's not all.

Orbiter:
This is good work, however I had to modify it to get it working and it worked only under certain conditions:

1) you must put in a Shebang into the first line ("#!/usr/bin/env python")
2) the package test is very good, it showed me what to do, however there is a spell error. Instead of
```pip instal requests``` it must be ```pip install requests``` (one more 'l')
3) the script runs only if it is started from the root of the project. If I start it from the bin subdirectory, it cannot find the test subdirectory. Please check how this is handled in bin/start.sh (using the line ```cd `dirname $0`/..```) and add something similar to your code

sevazhidkov:
Thanks, done.

Orbiter:
perfect! Thank  you!

PR #155, opened 2015-12-22
Made it shorter and Python2+Python3 compatible

yasoob:
Made it shorter and Python2+Python3 compatible

### DIFF
--- a/bin/search.py
+++ b/bin/search.py
@@ -1,25 +1,20 @@
 #!/usr/bin/env python
 """
-Requirements:
+    Requirements:
 
-requests (installation: pip install requests)
+    requests (installation: pip install requests)
 """
 
-import argparse
+from __future__ import print_function
 import requests
-import json
+import sys
 
-parser = argparse.ArgumentParser(description='Process the query input')
-parser.add_argument('query', type=str, help='Query to search for')
-args = parser.parse_args()
-query = args.query
+if len(sys.argv) < 2:
+    print('Please pass in the query which you want to search for')
+    sys.exit(-1)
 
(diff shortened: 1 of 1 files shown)


sudheesh001:
@yasoob How about having a look at the https://github.com/loklak/python-loklak-api#search-api

yasoob:
Sorry. I am doing it for Google Code-in :) And the task specifically mentioned to modify this file. Kindly correct me if I am wrong.

Orbiter:
This is actually much shorter. Why did you remove the "Requirements" comments? They are useful.

yasoob:
Hi @Orbiter. In my opinion the requirements should be listed in the ```requirements.txt``` file. That way you can easily install them with one command as well. If you don't have the required package installed then you will get an error while running this script which will tell you what you are missing. Therefore, I don't think there is any need for the "requirements" comment here.

However, we can add it if you want. It won't hurt anyone if it is there :+1:

Orbiter:
please add it in the same file.

yasoob:
I have added it. :+1:

sudheesh001:
@yasoob Can you squash this into one single commit ?

yasoob:
@sudheesh001 I am doing just that :sparkles: In the mean-time why don't you mentor me on [this](https://codein.withgoogle.com/dashboard/task-instances/5187189314945024/?sp-page=1) task?

yasoob:
I have squashed it. AND I have resubmitted the task for review as well. Would you mind taking another look at it?

Orbiter:
thank you!

PR #163, opened 2015-12-24
Add test for defined directory

PiotrKowalski:


bin/test_piotr.py
```diff
@@ -0,0 +1,8 @@
+import requests
+f = open('../test/queries/english_00.txt', 'r')
```
sudheesh001: Can you use this to read any of the files present in the `queries` directory, read each file and fire queries for each query in the file. That way its a generic script to fire anything in the `queries` directory.

yasoob: @sudheesh001 I have submitted a PR with exactly the kind of functionality you are talking about :+1:

bin/test_piotr.py
```diff
@@ -0,0 +1,27 @@
+
```
sudheesh001: The python `shebang` is missing here, Could you please add that?

sudheesh001:
@PiotrKowalski I've figured out the issue you were facing. The problem is that you're not converting the response to a `json()` , Requests by default encapsulates the json responses. So here is the work around
```python
for line in current_file:
		url = "http://localhost:9000/api/search.json?q=" +  line
		response = requests.get(url)
		x = response.json()
		try:
			print(x['statuses'])
		except Exception as e:
			print(e)
```

sudheesh001:
@PiotrKowalski Please squash everything into one single commit !

sudheesh001:
Please resolve conflicts and squash :smile:

PiotrKowalski:
I am trying to do it :P

Orbiter:
@sudheesh001 can you please test this and merge it, I am currently out of office. Thank you!

sudheesh001:
@Orbiter Tested :smile:

PR #173, opened 2015-12-31
Add document for thread dump API, fix errors in api.html

DengYiping:
for gci2015


DengYiping:
@sudheesh001 It is squashed.

PR #177, opened 2016-01-01
Create jars for distribution

cshubhamrao:


sudheesh001:
Is it really necessary to include an extra  `/>` space as `_/>` where `_` is the space ? Maybe you could just include the `<target>` for the `jar` as you've written. Could you please redo this with only the changes  necessary rather than changing the formatting. This looks exciting, looking forward to this.

cshubhamrao:
Ohk, maybe it was Eclipse's formatting, redoing the commit

sudheesh001:
You added the `.jar` file too, i don't really think that's necessary, we can have .jar releases if needed. Please remove that from the commits.

cshubhamrao:
.jar file requires the `html/` `conf/` and `lib/` dir also. Thats something to note. Releases will have to include those dirs also

sudheesh001:
Please squash all commits into one place.

cshubhamrao:
Commits squashed

Orbiter:
good work, I moved the jar file to a dist subdirectory

cshubhamrao:
It's not over yet. Here is what I think can be the method of distributing loklak:

* Modify `build.xml` further so that it creates a `release.zip` containing the following:
  * `conf/`, `lib/`,  `html/` directories.
  * Latest `loklak.jar` built by `ant`
* This `release.zip` will be uploaded to GitHub Releases.
* A python script (distributed as `.exe`) will do the following
 * Check presence of Java JRE, prompt user to download if absent.
 * Download the `release.zip` via GitHub API or other simpler means.
 * Extract the zip to a semi permanent location (My Documents?)
 * Run  the `.jar` 

Advantage is, only the python script will require tweaking, and same script can be used for other platforms too.

sudheesh001:
We can use the `git tag` feature to keep releases and the changelog between releases. Sounds good @cshubhamrao

PR #178, opened 2016-01-02
Fix problems in 2 files, add profiling script based on scala

DengYiping:
In order to use this profiling script, you should install the
latest version of scala and sbt(basically sbt will install everything
for you).
Then cd into loklak_server
```cd bin```
```sbt run```

Fixed files:
One file has format issue, one file has broken links.


DengYiping:
@mariobehling Here is a pull requests for GCI2015

PR #179, opened 2016-01-02
Added test cases for Hindi and English

shiftsayan:
Sourced from commonly tweeted and spoken words (including slang).


Orbiter:
oho, wow, Hindi test words! Thank you!

PR #181, opened 2016-01-03
Added 200 Common words 

shashank-sharma:
Added 200 more words (since there are few pull request with english_02.txt so I made this english_03.txt)


Orbiter:
great, thank you!

PR #185, opened 2016-01-06
Added search.py, fetches tweets from loklak.org API server

Boot-Error:
[GCI2015 challenge](https://codein.withgoogle.com/dashboard/task-instances/5886970718846976/?sp-page=1) that I have claimed

bin/search.py uses default python distribution module, and works for python2.X

### DIFF
--- a/bin/search.py
+++ b/bin/search.py
@@ -1,20 +1,21 @@
 #!/usr/bin/env python
-"""
-    Requirements:
 
-    requests (installation: pip install requests)
-"""
-
-from __future__ import print_function
-import requests
+import urllib
+import json
 import sys
 
-if len(sys.argv) < 2:
-    print('Please pass in the query which you want to search for')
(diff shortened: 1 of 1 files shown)


bin/search.py
```diff
@@ -1,20 +1,21 @@
...
+statuses = json.loads(fetchData).get("statuses")
+texts = [tweet.get("text") for tweet in statuses]
 
-query = sys.argv[-1]
-url = 'http://loklak.org/api/search.json'
-data = requests.get(url, params={'query': query})
-for status in data.json()['statuses']:
-    print(status['text'])
+for text in texts:
+	print "TEXT:",text
```
sudheesh001: `print text` should be enough in this situation.

sudheesh001:
Please squash into a single commit. @Boot-Error 
To know more about squashing you could have a look at : https://asciinema.org/a/11269

PR #191, opened 2016-01-07
Edit link : Link was not able to open

shashank-sharma:
After going through api.html page I found this example link at last. When I clicked it was not working here is the image: I removed localhost and putted loklak.org .And then it was working fine showing image [like this] (http://www.loklak.org/vis/markdown.png?text=hello%20world%0Dhello%20universe&color_text=000000&color_background=ffffff&padding=3)
screenshot from 2016-01-07 22 23 22

### DIFF
--- a/html/api.html
+++ b/html/api.html
@@ -1407,7 +1407,7 @@ added fuzzyness for marker computation
 
 <p>Here is an example :
 <ul class="request">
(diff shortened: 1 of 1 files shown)


Orbiter:
well the link works if you run the server yourself. In fact all links in the api should refer to localhost:9000, not to loklak.org. Please consider loklak.org just as a 'demo' instance of loklak and not as 'the' service. As a peer-to-peer software, you should always consider to run the software yourself.

PR #196, opened 2016-01-08
first

nolik2000:


nolik2000:
https://github.com/nolik2000/loklak_server

sudheesh001:
@nolik2000 What's the change made in the min file, ideally any css changes should not be in the default libraries, it should be outside it, overriding the css where ever necessary.

PR #197, opened 2016-01-09
Added querybrowser

anshuman73:
As the [task description in GCI](https://codein.withgoogle.com/tasks/6304774081216512/?sp-is_beginner=False) stated that this page will be used in the loklak front end, I have used angular.js only to call for requests. If hosted on different server, this will not work due to cross domain error. I haven't been able to test it for the same reason. Otherwise, it should work fine

### DIFF
--- a/html/querybrowser.html
+++ b/html/querybrowser.html
@@ -1,0 +1,98 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


Orbiter:
its not working, my firefox console reports an error:
"ReferenceError: angular is not defined" in line 81.
I believe there is something missing. Overall, this looks good, but please fix this to complete the task.

anshuman73:
ahhh
That means angular is not getting loaded. Alright, give me a second

PR #198, opened 2016-01-09
Fixed Angular Problem

anshuman73:
Fixed the problem of angular.js not loading. Now it fetches it directly from the google's servers

### DIFF
--- a/html/querybrowser.html
+++ b/html/querybrowser.html
@@ -93,6 +93,6 @@
     <script src="js/docs.min.js"></script>
     <!-- IE10 viewport hack for Surface/desktop Windows 8 bug -->
     <script src="js/ie10-viewport-bug-workaround.js"></script>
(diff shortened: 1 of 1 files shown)


anshuman73:
fixed issues with #197

Orbiter:
This does not work either. I tried to move the original angular import to the head section of the html document and it worked, but only after I made the API request into a localhost request.
You can see the changes in https://github.com/loklak/loklak_server/commit/20942e7bd259ef67fa8988ad09dd66111be84b24

It works pretty well now, I consider this task as completed. Thank you very much for this code!

I will now move your work into the apps folder, so your api visualizer becomes a loklak app.

anshuman73:
ahh. Well, like I told at the very beginning, angular doesn't really work well with get requests if not on same domain.

And funny moving the request into makes it work.

well, thank you for the task, though I didn't even claim it ;p

PR #199, opened 2016-01-10
english_04.txt

shiftsayan:
I have created 200 test cases which contain names of various brands. This can be very beneficial because the names of brands remain constant over various languages and regions!


Orbiter:
very good idea! Thank you!

PR #203, opened 2016-01-10
Remove annoying white line at the bottom!

darrylyeo:
Removed a line from a CSS rule that added `padding-bottom` to the `<body>`, causing a white line to appear at the bottom of the page.

A suggestion would be to add a black background to the `<body>` as well, because some operating systems allow you to scroll past the top and bottom before snapping back.


Orbiter:
yeah! very good, thank you!

PR #206, opened 2016-01-11
Add swapfile to Google and AWS instance.

DengYiping:
The default AWS instance does not have swapfile, my server was down couple times due to lack of memory. I simply add 4G of swapfile in this case instead of upgrade the instance(which will cause you some money since we have used the largest free tier instance already)


DengYiping:
This patch is necessary especially for the free tier instance in AWS. They only have 1G of memory, and it is not enough to run loklak-server.

sudheesh001:
Looks right to me ! Good work @DengYiping

PR #208, opened 2016-01-11
Add instruction for Loklak installation in DigitalOcean using Docker

sevazhidkov:


Orbiter:
Very well, thank you!

sudheesh001:
We could also use the VPS droplet and set it up the same way as on a VM. Good work @sevazhidkov

PR #211, opened 2016-01-13
Improved setup instructions for AWS

JLeow00:


Orbiter:
thank you!

DengYiping:
@JLeow00 Hello! Do you use a micro instance? Seem like my server on AWS is not stable and it crash couple time, do you face the same issue?

JLeow00:
@DengYiping Yes I used a micro instance, but so far my server hasn't
crashed

DengYiping:
@JLeow00 I have set up my AWS instance as backend, seem like I'm crawling too fast.

PR #230, opened 2016-01-19
added query input in histogram app 

yasoob:


Orbiter:
Thank you! I tested this, the page now provides an input box, but when I do a search I get the dialog box with "Kindly provide a query to search for!" - and no graphics. Tested with firefox.

yasoob:
That is intentional. It gives that alert when there is no query in the
input field. If you want me to change it then i can change it.

I have got two ideas:

1. Automatically fill in the query input with loklak or fossasia on page
load.

2. Use loklak or fossasia as default input value when the input box is
empty and the user has clicked on plot.

Which approach do you like better?

DengYiping:
@yasoob I think it is a bug in this web app. I just fixed it.

yasoob:
@DengYiping Let me verify.

yasoob:
Hey @Orbiter I just realized the issue. I have updated the code and submitted a [pull request](https://github.com/loklak/loklak_server/pull/233).

DengYiping:
I've submit a pull request to fix the problem @yasoob . And I refactored your code.

yasoob:
Hey @DengYiping Sorry but I didn't see your PR before. I will close mine. :+1:

PR #236, opened 2016-01-21
Added scroll-spy on 'about.html' for better scrolling

rkakkar7:
### CHANGES TO ABOUT.HTML
* Added Twitter Bootstrap's Scrollspy for a better scrolling experience.
* Packed the nav-sidebar in a `<div>` with *id='navSidebar'* so that it can be targeted by the *data-target* in the *body* tag.
* Grouped the sections using `<section>`.
* Removed the *'id'* attribute from `<br>` and added to the newly made `<section>`.
* Removed the useless `</span>` tags in *nav-sidebar*.

**Usage**
It shows the current section in the browser's view by highlighting the corresponding link in the *nav-sidebar* by dynamically updating upon scrolling.

### CHANGES TO DASHBOARD.CSS
* Added a new style for the *nav-sidebar* so that it can highlight the *active* class in the *nav-sidebar*

### SCREENSHOTS

P.S. - Please ping me if you would like it to be implemented for all the pages.
Also if you could provide me with a link to the issue tracker so that I can remain updated about the current issues. Thanks.

### DIFF
--- a/html/about.html
+++ b/html/about.html
@@ -23,7 +23,7 @@
     <![endif]-->
   </head>
 
-  <body>
+  <body data-spy="scroll" data-target="#navSidebar" data-offset="50">
 
     <nav class="navbar navbar-inverse navbar-fixed-top">
       <div class="container-fluid">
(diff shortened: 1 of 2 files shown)


Orbiter:
Hi
thats pretty useful!, Thank you.
However, there is now a problem: if you click on the menu items on the left side, the headline of that section is not visible; it's above the content view. That was the reason I added the ```<br/><br/>``` after the section marker. Now these line breaks are before the marker. I do not know how to handle this now, maybe we must drop the extra line breaks and add a css style change to make the section headlines visible again. Please try to solve this problem.

PR #238, opened 2016-01-22
Add my search page

ducphanduyagentp:
I've just added my search page in /html/apps/.


ducphanduyagentp:
I haven't seen my app on the Apps section of loklak.org? Is this an issue? The app appears normally on my machine with the exact files.

PR #239, opened 2016-01-23
GCI 2015 tweet task for Fossasia

peguin40:
Submit of coding task result for GCI 2015. Username in GCI: Ang Swee Chow

### DIFF
--- a/bin/search_fossasia.py
+++ b/bin/search_fossasia.py
@@ -1,0 +1,27 @@
+#Returns all tweet text lines related to fossasia
+import urllib
+from urllib import request
+import html
+import json
+import sys
+
+query = "http://loklak.org/api/search.json?q=fossasia"
+
+#get file and decode
+try:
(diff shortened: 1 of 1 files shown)


bin/search_fossasia.py
```diff
@@ -0,0 +1,27 @@
+#Returns all tweet text lines related to fossasia
+import urllib
+from urllib import request
+import html
+import json
+import sys
+
+query = "http://loklak.org/api/search.json?q=fossasia"
```
sudheesh001: Can you make the `q` parameter be a variable instead of hardcoding it to `fossasia` ? Take the `argv` values and you can do this.

sudheesh001:
Please squash commits into one single commit.

peguin40:
How do you do that?

sevazhidkov:
@peguin40 http://git-scm.com/docs/git-rebase#_interactive_mode may be helps you :)

Issue #265, opened 2016-02-26
Server does not shutdown


shivamMg:
Even after issuing `bin/stop.sh` at the terminal the server keeps on running at `localhost:9000`. A message that loklak has been terminated is displayed but service is still available at localhost.

```
$ bin/start.sh 
starting loklak
loklak server started at port 9000, open your browser at http://localhost:9000

$ bin/stop.sh 
shut down of loklak
loklak has been terminated

$ bin/stop.sh 
shut down of loklak
cat: loklak.pid: No such file or directory
loklak has been terminated
```

sudheesh001:
It's probably just the cache, did you try opening up a private browsing with the same localhost url and seeing if it works ?

shivamMg:
It's not the cache. I opened it in private browsing and it's running.

SilentFlame:
@shivamMg  Initially I got the same error.
but when I used the private window opening as mentioned by @sudheesh001 , It works fine for me though.

SilentFlame:
@sudheesh001 can we do something regarding the cache thing here?

shivamMg:
Ok, got it. Whenever we run `bin/start.sh`, it starts a java server. The PID of this server is stored at `data/loklak.pid`. Now when we run `bin/stop.sh` it takes the PID of the existing server instance from the file and kills it. The problem appears when you start the server with `bin/start.sh` but do not stop an existing instance of it. If this happens, `bin/start.sh` overrides the existing PID with the PID of the new instance, and if you stop it with `bin/stop.sh` now, the new server instance is stopped but the older PID is now lost. So if you keep on executing `bin/start.sh` you will have multiple server instances with lost PIDs except the latest one.

Solution is simple. We can check if there is already a running instance of loklak, remove it, if it exists, and start a new instance again.

shivamMg:
@sudheesh001 I have fixed the bug and will publish the changes. I have an older PR https://github.com/loklak/loklak_server/pull/266 if you can review it.

Orbiter:
The right behavior should be
- do not start a new process if one is running already
- do not overwrite the pid
- show a message that the process is already running

SilentFlame:
Either we are done here with the server, or we can create a new file for restart of the server.
which first quits the running server and then starts the new server.
@sudheesh001 @Orbiter @shivamMg 

what you guys say?

smokingwheels:
@SilentFlame @sudheesh001 @Orbiter @shivamMg
[...]

PR #269, opened 2016-02-27
Fix bug #265: Shutdown existing server

shivamMg:
Shutdown existing server before starting a new one.

### DIFF
--- a/bin/start.sh
+++ b/bin/start.sh
@@ -1,4 +1,15 @@
 #!/usr/bin/env sh
+
+# Check for an existing instance of server
+# If it exists, shut it down
+filepath="`dirname $0`/../data/loklak.pid"
+if [ -f $filepath ]; then
+ echo "shutting down existing loklak server"
+ kill `cat $filepath` 2>/dev/null
+ if [ $? -eq 0 ]; then while [ -f $filepath ]; do sleep 1; done fi;
+ rm -f $filepath 2>/dev/null
+fi
+
 cd `dirname $0`/..
 mkdir -p data
 


smokingwheels:
Is it possible to check the log for ports in use?
I had the problem where the elastic search had crashed and was tieing down the other ports, port 9000 was ok

sudheesh001:
@shivamMg anything to say about the query put up by @smokingwheels ?
This looks good to me, but I'd let @Orbiter or @smokingwheels test this and merge.

shivamMg:
Noted. 

`bin/start.sh` will first check for an existing server. If it exists it'll just display a message ("loklak already running") and exit.

I didn't notice other scripts before. The script `bin/checkalive.sh` seems to do the same, check if the server is up by making a call at localhost. But shouldn't the correct way to do this be checking for PID in `data/loklak.pid`. I get it that it also checks for "loklak" string from the "index.html" page, but still, there could be other servers running at localhost. 

Plus `bin/checkalive.sh` is doing what a restart.sh should be doing, check for conditions, kill the process and then starting it again. So you can't use `bin/checkalive.sh` in `bin/start.sh`, since it'll be doing a restart itself.

I think checkalive.sh should (only) have a function `is_alive` that just checks if the server is up. We can then source this script and use the function. A separate restart.sh should be created that essentially does what checkalive.sh is doing, i.e. killing and then starting the server.

@Orbiter What do you think?

shivamMg:
@smokingwheels Sorry for the late reply. No, I checked the logs and they don't store the ports in use. Aren't the ports in use defined in the `conf/config.properties`, or in `customized_config.properties` if you have altered them. Or is there something I'm missing.

Orbiter:
thank you for your code, however the behaviour is better if we do not kill existing processes just if you start another. The rule to take care of it was implemented in https://github.com/loklak/loklak_server/commit/bfa134f0b295f1fe8a53ad9f74cdefc0b1da9f83

PR #273, opened 2016-03-05
overwriting of running process handled

SilentFlame:
@Orbiter  it handles all the three of your mentioned points for right behavior of the server in the #265 
@sudheesh001 ,
have a review, and issue #265 can be closed after this merge.

### DIFF
--- a/bin/start.sh
+++ b/bin/start.sh
@@ -1,6 +1,12 @@
 #!/usr/bin/env sh
 cd `dirname $0`/..
 mkdir -p data
+#to not allow process to overwrite the already running one.
+x=`cat data/loklak.pid 2>/dev/null` && ps -A | egrep "^\ $x" >/dev/null 2>/dev/null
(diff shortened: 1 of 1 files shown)


SilentFlame:
@sudheesh001 can you please review it and merge it. 
as need your suggestion to make a restart.sh file or not.

Orbiter:
oh wow thank you!

SilentFlame:
@Orbiter  @sudheesh001 
I was thinking to make a `restart.sh` to quit and start the server again in one go, as it's annoying at the time of editing the code and running the `start` and `stop` again and again. It would be great to have that.
what's your say?

Orbiter:
yeah thats useful but it should just be a stop.sh;start.sh

or not?

SilentFlame:
ya that's what it is,
we can make a file with comments in between. 
`process with pid:XXXX shutting down, new process started on port : 9000`
something like this.

PR #292, opened 2016-04-14
User profile app url fix 

daminisatya:


sudheesh001:
Looks like there's a small problem here @daminisatya , this already contains the previous commit that was merged and hence contains 2 commits. Since the previous pull request is already merged. I think what you should do for this is firstly
```sh
git checkout master
git reset master --hard 
git branch -D userAppURLfix
git pull upstream master
git checkout -b upserAppURLfix
# Redo the commit and mention the fix url in the commit message
git commit -m "Fixes #291 ......."
```

daminisatya:
@sudheesh001 done

sudheesh001:
@daminisatya Looks great 👍 Thank you

PR #304, opened 2016-04-24
Updated the GeoCode JSON format in the API docs

daminisatya:


sudheesh001:
@daminisatya Looks good, thanks for updating the JSON structures for the GeoJSON, The sample query URL given in the paragraph above isn't a link and isn't formatted correctly, can you make that a link like how we have in `search.json`. This should then be good for merge.

PR #308, opened 2016-04-29
Updated the suggest API JSON structure in Docs

daminisatya:


sudheesh001:
@daminisatya r+

This looks good, I don't see a problem with this, also `hits` is now a part of the `search_metadata`. Merging this.

PR #310, opened 2016-04-30
[Heroku_Installation_Guide]Insert the missing step before creating a new app

jig08:
Insert the missing step of moving into the `loklak_server` repo before creating a new app


sudheesh001:
@jig08 r+ This seems better. Thanks, merging this.

PR #320, opened 2016-05-05
Added autorecovery for bluemix

shivenmian:
I'd faced a problem that my bluemix server used to go down every week or so, probably because of its high memory usage / activity. Using autorecovery restarts the container group when it goes down.


sudheesh001:
You can also click the autorecover option while using the container group right ?

PR #321, opened 2016-05-05
Inserting Docker One-Click Deployment Button 

jig08:
[Demo Link](https://github.com/jig08/loklak_server/blob/master/README.md)

PR #322, opened 2016-05-07
Inserting Docker One-Click-Deployment Button

jig08:
[Demo Link](https://github.com/jig08/loklak_server/blob/master/README.md)

This is an update from previous PR https://github.com/loklak/loklak_server/pull/321 regarding the same issue


sudheesh001:
conditional review+ endpoints still not public

PR #324, opened 2016-05-08
Added Tweet badge to README.md

daminisatya:
I thought it would be awesome to add a tweet badge when clicked would auto tweet on to the twitter from the user.


jig08:
@daminisatya Great.
Wouldn't it be nice to insert the `@loklak_app` or `@lklknt`handle mention in the default tweet text ?  :)

daminisatya:
@jig08 Good one! Will add it

daminisatya:
#325

Issue #323, opened 2016-05-07
Docker cloud deployment endpoint not public


sudheesh001:

sudheesh001:
Found the bug, the bug was in the deployment methods, for loklak to run, we actually need 4G of memory but the docker containers can run on 1G RAM also. This minimal fails on Azure node with A0 because they offer only 0.75G because of which the build works but it becomes unable to run the application as a whole and keeps crashing even though the port 80 is exposed in the stack file.

To overcome this problem we always need the docker nodes which are > 1GB RAM so that the application containers can be deployed and hosted over there. The minimal machine doesn't get this up and running.

Here's a running instance on a 2 Node docker cluster with 1 container of loklak image on it.

http://loklak-1.loklak.990befd4.cont.dockerapp.io:32768/

Maybe we should record this in the documentation for others who are planning to deploy docker instances. Also please export the port 443 along with 80

sudheesh001:
Same issue with AWS and Digital ocean, please take minimal nodes 2 so that 1 can be dedicated to disk and other to operating system. Also choose a slightly larger machine to run this >1.5G for safety and do not choose a machine lower than that during deployment.

PR #328, opened 2016-05-08
Exposing port 443 to make Docker Cloud deployment endpoint public

jig08:
@sudheesh001 Changes done as proposed wrt Issue #323.

Earlier, it showed `32768->80/tcp` as the Published Port and `443/tcp` as the private port.
Now it is showing `32769->443/tcp` as the Published Port and `80/tcp` as the private port.

### DIFF
--- a/docker-cloud.yml
+++ b/docker-cloud.yml
@@ -2,4 +2,4 @@ web:
   name: "loklak"
   image: mariobehling/loklak
   ports:
-    - "80"
+    - "443"


docker-cloud.yml
```diff
@@ -2,4 +2,4 @@ web:
   name: "loklak"
   image: mariobehling/loklak
   ports:
-    - "80"
+    - "443"
```
sudheesh001: Please don't delete the `80` instead add 2 port values, that's what was expected
```
- "80"
- "443"
```

jig08:
@sudheesh001 Oh, now I get it.
Will do the required changes :)

jig08:
@sudheesh001 Here are the new changes #334

PR #329, opened 2016-05-09
Fixed Bluemix Deployment

shivenmian:
I checked the Bluemix docs, and so far I've figured this out:

1. The buildpack url doesn't go into the env of the manifest.yml, [it has to be specified separately](https://console.ng.bluemix.net/docs/manageapps/depapps.html) as ```buildpack: BUILDPACK_URL```. So I don't think the deployment is even noticing the buildpack right now. I checked the format from [here](https://console.ng.bluemix.net/docs/manageapps/depapps.html#appmanifest)

2. Probably pipeline.yml is needed because the app uses containers.

I added the blocked label. @sudheesh001 @jig08 @Orbiter @daminisatya how do I test it?

PS: Did the assignment by mistake, sorry.

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,9 +1,8 @@
----
-  name: "Loklak Server"
+--- 
+- name: "Loklak Server"
+  buildpack: "https://github.com/aneeshd16/heroku-buildpack-ant-loklak.git"
   description: "Distributed Tweet Search Server"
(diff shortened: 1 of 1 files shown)


sudheesh001:
The only way we can test it is by merging because it takes the referrer url and the script updates.

PR #333, opened 2016-05-10
Added to Readme

shivenmian:
Added two more badges, which I believe will be useful for contributors. We need to resolve as many of those issues as we can. @Orbiter @sudheesh001 @jig08 @daminisatya


jig08:
`review+`  Looks good to me :+1: Can be merged.

daminisatya:
Good one! Merging it.

PR #339, opened 2016-05-13
Setup -Xmx from environment

Soulou:
In environments like Scalingo or Heroku, it would be much more easy to configure the Xmx from en environment variable.

Then, you can configure it easily accordingly to the type of container/dyno you're using.

Variable name is "ENVXmx" to fit other variables naming scheme

### DIFF
--- a/bin/heroku_start.sh
+++ b/bin/heroku_start.sh
@@ -24,7 +24,7 @@ CLASSPATH=".:./classes/:$CLASSPATH"
 
 cmdline="java";
 
-if [ -n "$ENVXmx" ] ; then cmdline="cmdline -Xmx $ENVXmx";
+if [ -n "$ENVXmx" ] ; then cmdline="$cmdline -Xmx$ENVXmx";
 elif [ -n "$CUSTOMXmx" ]; then cmdline="$cmdline -Xmx$CUSTOMXmx";
 elif [ -n "$DFAULTXmx" ]; then cmdline="$cmdline -Xmx$DFAULTXmx";
 fi
--- a/bin/heroku_start.sh
+++ b/bin/heroku_start.sh
@@ -24,7 +24,8 @@ CLASSPATH=".:./classes/:$CLASSPATH"
 
 cmdline="java";
 
-if [ -n "$CUSTOMXmx" ]; then cmdline="$cmdline -Xmx$CUSTOMXmx";
+if [ -n "$ENVXmx" ] ; then cmdline="cmdline -Xmx $ENVXmx";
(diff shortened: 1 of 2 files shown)


shivenmian:
@Soulou Should this change also be made in bin/start.sh along with bin/heroku_start.sh?

Also, to fix the crashing of loklak on Scalingo, I'd spoken to Leo and he said that Scalingo and Bluemix supports max 512m, so Xmx in ```config/config.properties``` should be set at 2/3 of that i.e 350m. However, for Heroku, 800m is max.

The app actually crashes on Scalingo, Heroku and Bluemix within 1-7 days, and swapping is causing the poor performance (the Xmx as of now in ```config/config.properties``` is set at 600m and $DFAULTXmx is set at 800m in ```bin/start.sh``` and ```bin/heroku_start.sh``` both).

So to accomodate all three platforms, what value should be set for $DFAULTXmx in ```bin/start.sh``` and ```bin/heroku_start.sh``` and Xmx in ```config/config.properties```?

Soulou:
Actually, it's 512m is not the max, it's the size of the standard resource 1 dyno at Heroku or 1 container M at Scalingo. If you choose to have larger resources, you should be able to increase the size of Xmx dynamically, that's why I think this pull request is useful.

Having to modify a file each time the environment is changing is not a great way to do, it does not respect the [12factor rules](http://12factor.net). Most PaaS expect the apps to be based on thes rules.

tldr There is no one single value to set for `bin/start.sh`, `bin/heroku_start.sh` or `config/config.properties`, it should be configurable.

350m is ok for a 512MB server
600m is ok for a 1GB server
1200m is ok for a 2GB server
etc.

shivenmian:
Oh okay, I understand. It does make sense.

PS: May sound naive, but why have you not assigned ```$ENVXmx``` and used it directly? Or does the compiler pick it up automatically?

Soulou:
Actually, I've done it this way to fit the ideas behind the scripts, if it would depend of me, I would rewrite the scripts completely :-)

shivenmian:
@Soulou Oh okay so this script can't be merged yet? We need to assign ```$ENVXmx``` / rewrite the script?

Soulou:
No it's ok this PR can be merged, I was just saying that the script is a bit messy and could be rewritten, if I've time, I'll do it :-)

It just adds the ENVXmx environment variable which can override all the others Xmx definition (configuration etc.)

So it becomes much more flexible.

shivenmian:
@sudheesh001 Merge this?

Issue #318, opened 2016-05-03
Bluemix one click deploy failing in deploy stage build successful


daminisatya:
Same errors as in : https://github.com/loklak/loklak_server/pull/316#issuecomment-216429691

jig08:
I believe the reason could be missing `declared services:` in the `manifest.yml` file.
We need to mention the optional services that are expected to be set up before the app is deployed.
Could you please help me list 'em out ?

@Orbiter @daminisatya @sudheesh001

Orbiter:
I cannot find a hint about which service is missing, do you have more error logs?

jig08:
@Orbiter My `IBM Bluemix` free trial period \*just* expired :cry: 
Now the `Project Creation` only fails at the first step.

sudheesh001:
@jig08 Since the build works fine I don't think there's a problem with any of the dependencies here, the issue is in running the start as far as i can guess.

jig08:
@sudheesh001 I tried exploring a bit, but no luck so far.
Will read more about stage's run-time information, maybe.

sudheesh001:
@jig08 You might need a `.pipeline.yml` file i guess for the stages

jig08:
@sudheesh001 Probably yes. We need to add the `pipeline.yml` in the `.bluemix` directory.

Should we create one manually ? Or if you have a valid Bluemix account (mine just expired a few days back), you could generate one from the existing project there.

Ref [Creating a Deploy to Bluemix button](https://console.ng.bluemix.net/docs/develop/deploy_button.html)

sudheesh001:
Build succeeded but deploy failed again @shivenmian

shivenmian:
Logs?

shivenmian:
Okay @sudheesh001 @jig08 

I checked the logs. There was a problem in the buildpack syntax (we were missing the applications: tag). I will add it. But still when I added it locally it wouldn't deploy, because of the same staging error.

I'm definitely sure we need a pipeline.yml because that's what defines the stages. I'll work on it and submit a PR shortly.

jig08:
@shivenmian Yes, as mentioned https://github.com/loklak/loklak_server/issues/318#issuecomment-217609641 we could create `pipeline.yml` either manually or generate one from the existing project in Bluemix.

If you have a valid account, pls go ahead to include an auto-generated one.

Reference: [Creating a Deploy to Bluemix button](https://console.ng.bluemix.net/docs/develop)/deploy_button.html

[...]

PR #341, opened 2016-05-13
Added pipeline.yml, edited manifest to use modified buildpack

leonmak:
My stab at fixing #318 for now. To get the pipeline.yml file, after the deploy failed, I looked at the default build step config and removed the default "output" archive directory, then there were some errors like missing java path so I modified the buildpack a bit, search works but I'm not sure if there are other dependencies missing.

### DIFF
--- a/.bluemix/pipeline.yml
+++ b/.bluemix/pipeline.yml
@@ -1,0 +1,37 @@
+---
+stages:
+- name: Build Stage
+  inputs:
+  - type: git
+    branch: master
+  triggers:
+  - type: commit
+  jobs:
+  - name: Build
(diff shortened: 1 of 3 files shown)


sudheesh001:
@leonmak This is awesome, I pulled from your fork and it seems to be awesome, also it registers as a CFApp even after using a container image.

sudheesh001:
http://loklak-server-dispensary-92.mybluemix.net/

screen shot 2016-05-14 at 9 07 50 am

screen shot 2016-05-14 at 9 07 39 am

screen shot 2016-05-14 at 9 08 25 am

PR #350, opened 2016-05-14
Remove deprecated and unsecure ssl client connection

treba123:
Java now prefers to use system defaults, therefore custom certificates should still work by importing them systemwide


treba123:
Had to move it to #351

PR #351, opened 2016-05-14
Remove deprecated and unsecure ssl client connection

treba123:
Java now prefers to use system defaults, therefore custom certificates should still work by importing them systemwide

### DIFF
--- a/src/org/loklak/http/ClientConnection.java
+++ b/src/org/loklak/http/ClientConnection.java
@@ -27,12 +27,6 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.net.UnknownHostException;
-import java.security.KeyManagementException;
-import java.security.KeyStoreException;
-import java.security.NoSuchAlgorithmException;
-import java.security.SecureRandom;
-import java.security.cert.CertificateException;
-import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


treba123:
Ok sorry for the mess, I first needed to fix an upstream build problem and somehow missed something first. Works now

Orbiter:
well I am not sure if this is good for this project. Lets look at the purpose of ssl connections:
- provide encryption on content
- provide server authentication
What we need here is content encryption. What we don't need is server authentication because we want to communicate between anonymous installations. This change will mean that we cannot establish a communication between ssl-encrypted peers before we have forced them to authenticate.

However, we could start to distinguish two different trust levels for communication:
http - low trust
https with self-signed certificates - medium trust
https with valid certificates - higher trust

This could be an attribute in the context of https://github.com/loklak/loklak_server/issues/353

I therefore would like to use the untrusted connections by default but would also have the opportunity to use trusted as well.
Can you please modfy your pull request in such a way that the new client connection becomes an option, not the default

treba123:
@Orbiter I see your point, but one second:
- right now, self signed certificates are still possible (by importing the certificate)
- encryption without authentication is not really encryption as it doesn't prevent man-in-the-middle attacks
- If I get it right, connections are usually build up to the backend server and twitter, both satic and known parties
Do we have scenarios were we want to connect dynamically to unknown parties?

Anyway I can make a config setting to toggle authentication enforcement and set it to false by default

Orbiter:
- "encryption without authentication is not really encryption" - thats true!
- "scenarios were we want to connect dynamically to unknown parties" - yes, our own peers.
Ok we should distinguish this:
- force authentication for servers which should have non self-signed, trusted certificates
- lazy authentication for other servers, with the option to force here as well.

Issue #348, opened 2016-05-14
Data not collected - @user disappears


mariobehling:
![screenshot from 2016-05-14 08 13 29](https://cloud.githubusercontent.com/assets/1583873/15266620/f9223d68-19ab-11e6-9116-b9cce1193977.png)

![screenshot from 2016-05-14 08 14 16](https://cloud.githubusercontent.com/assets/1583873/15266622/fbad0b44-19ab-11e6-99a2-2237dafd8d16.png)

smokingwheels:
Not sure if it related but when you run your own server you get more results for the same quire and frequency as compared to loklak.org. The loss is about 75% when using loklak.org from the semiprivate test I have run.
Does Loklak.org have any forward peers?
Some TCP packets splatted somewhere by someone using there mobile phone??

leonmak:
I think it is because of the uploaded picture, seems to change the text and links

leonmak:
This seems to happen only when there are uploaded pictures

PR #354, opened 2016-05-17
Changed links regex to account for links ending with )

leonmak:
for eg tweet was : https://en.wikipedia.org/wiki/Joshua_Greene_(psychologist)
tweet becomes:
screenshot 2016-05-17 19 22 42
screenshot 2016-05-17 19 23 58

changed regex so it becomes
screenshot 2016-05-17 19 25 45


leonmak:
to fix #348, seems like the scrapper's regex pattern was too generous so it covered previous tags

PR #365, opened 2016-05-18
Fixed Boilerplate

shivenmian:
The boilerplate index.html as of now does not work. The sections in the menu, on clicking, don't lead anywhere and the theme (css) was not applying (it was all whitespace). This was because the directories used in the index.html (i.e the css, js, images directories etc) were not being referred to (boilerplate is in ```html/apps/boilerplate/index.html``` and css is in ```html/css``` so you can't simply refer to css using /css/loklak.css, it was not working). The path to those files in ```boilerplate/index.html``` is incorrect. It is fixed by adding ```../..```. 

@Orbiter I'm not sure if this is necessary as a commit, but I figured out this is why I could not use the boilerplate earlier, because the right files were not being referred to. I have fixed it. You don't need to copy any files from the CSS / JS directories to the new app directories now, you can simply copy the boilerplate and the default theme will come. This is the result of the commit:

screen shot 2016-05-18 at 7 58 04 pm

### DIFF
--- a/html/apps/boilerplate/index.html
+++ b/html/apps/boilerplate/index.html
@@ -7,20 +7,20 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="Content-Language" content="en, english"/>
     <meta name="author" content="@0rb1t3r"> <!-- CHANGE THIS -->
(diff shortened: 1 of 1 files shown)


Orbiter:
ok great thank you!

PR #369, opened 2016-05-19
Added Favicon to all the apps

daminisatya:


daminisatya:
Merging it since it is a small change.

PR #372, opened 2016-05-19
Regex fixes for text, links, mention fields for search.json, & added spaces before links

leonmak:
#348 occurs because of the scraper regex pattern, which replaces previous tags, so for eg when there is an uploaded photo or link, the mention is removed before it can be extracted as it has the anchor tag also. Adding a space before uploaded photo url in the text field will allow it to be extracted into the links field.

### DIFF
--- a/src/org/loklak/harvester/TwitterScraper.java
+++ b/src/org/loklak/harvester/TwitterScraper.java
@@ -367,10 +367,10 @@ public class TwitterScraper {
     }
     
 
-    final static Pattern hashtag_pattern = Pattern.compile("<a .*?href=\"(.*?)\".*?class=\"twitter-hashtag.*?\".*?><s>#</s><b>(.*?)</b></a>");
-    final static Pattern timeline_link_pattern = Pattern.compile("<a .*?href=\"(.*?)\".*?data-expanded-url=\"(.*?)\".*?twitter-timeline-link.*?title=\"(.*?)\".*?>.*?</a>");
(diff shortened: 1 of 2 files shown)


src/org/loklak/harvester/TwitterScraper.java
```diff
@@ -520,8 +520,8 @@ public static String unshorten(String text) {
             try {
                 Matcher m = timeline_embed_pattern.matcher(text);
                 if (m.find()) {
-                    String shorturl = RedirectUnshortener.unShorten(m.group(2));
-                    text = m.replaceFirst("https://pic.twitter.com/" + shorturl + " ");
+                    String shorturl = RedirectUnshortener.unShorten(m.group(1));
+                    text = m.replaceFirst(" https://pic.twitter.com/" + shorturl + " ");
```
sudheesh001: Is it needed to `replaceFirst()` with the initial ` ` characters ? Would it not be possible to trim the replacement when Matching the required pattern with the matcher ?

leonmak: The space is to prevent the case when the user uploads a photo. the pic.twitter.com link is joined with the text, this also causes the link field to be joined. For eg:
screenshot 2016-05-22 09 34 00
screenshot 2016-05-22 09 34 07
For me the text is still trimmed in the end. I'll check again and squash the commits too.

Orbiter:
sorry for not checking this earlier. This contains a lot of changes. Did you test this extensively?

leonmak:
I've tested quite a few combinations. There is just one issue of tweets with line breaks not being processed, after the line break. I'll see if I can fix that then I'll submit again

PR #384, opened 2016-05-20
Save client connection

treba123:
This is a big rewrite of the client connection. 
The first commits clean up the code of the ClientConnection and make them work well (actually, the old code did not accept certificates for wrong names, as it was supposed to do)

The later ones implement different download options for peers and non peers as suggested in #351 
The config has now three options for httpsclient.trustselfsignedcerts:
- none: never trust self signed certificates
- peers (default): allow self signed certificates for peers, not for other sources
- all: always accept all certificates

This comes with a warning: by making peers the default, everyone behind a ssl-proxy has either to change the config setting or import the proxy cert to her system. I think it's fair to be a little bit more "secure by default"

Whenever downloading from peers now, we have to use `downloadPeers` instead of `download` now.
`download` assumes authentication, but can be flagged to be opportunistic
Pushes assume peers by default, but can explicitly flagged to not do it.
If this is confusing, I can make it more similar.

I hope I got all download calls to peers changed, but ther're some doGet calls I didn't rightly understand. They still use the non peer download.

@Orbiter can you have a look on this later?


Orbiter:
all authentication/authorization rules should be implemented in https://github.com/loklak/loklak_server/blob/master/src/org/loklak/server/AbstractAPIHandler.java which handles this for all servlets. They propagate their authorization rules using https://github.com/loklak/loklak_server/blob/master/src/org/loklak/server/APIHandler.java#L32

I know that this is still an implementation stub. Lets see how we can merge this. I’m afraid your code needs ann adoption to that strategy.

treba123:
Ah ok. Right now we have code using the download fuctionality independently of AbstractAPIHandler at different places, even in DAO init.

Do I get it right that you
1. want all downloads to be called through AbstractAPIHandler
2. want AbstractAPIHandler do decide whether a url points to a peer and can therefor done over an unsave connection?

1 might be a bit of a hassle at the moment (or unnessecary for this pull request, it could be done in a follow up to make the all the loklak code comply with AbstractAPIHandler), but 2. seems quite important indeed.
Would 2. be enough to complete the pull in the moment? I'd like to avoid to make this pull request even bigger, as it was mainly intended to fix certificate checking and code cleanup @Orbiter

Issue #387, opened 2016-05-21
AppList matrix should link to each app


sudheesh001:

daminisatya:
@sudheesh001 I can fix this, since it is a small change with got missed out.

daminisatya:
#356

PR #388, opened 2016-05-21
Added Link to the apps page

daminisatya:


daminisatya:
#387

html/apps/applist/index.html
```diff
@@ -71,6 +71,7 @@
                   <div class="hidden">
                     <div class="caption">
                       <div class="centered">
+                        <p><a href='../../apps/{{ app.name }}'>{{ app.headline }}</a></p>
```
sudheesh001: It's okay to just href this link to `"/apps/{{ app.name }}"` instead of `../../`

shivenmian: That wouldn't work because the file is in the applist directory, so you can't access ```/apps/``` directly.

Although you could have written ```../'{{ app.name }}'```.

daminisatya:
@shivenmian This would work **/apps/{{ app.name }}** Since the apps page is being retrieved on to the main index.html outside the applist

daminisatya:
@shivenmian **../{{app.name}}** is fine too.

PR #389, opened 2016-05-21
Updated the apps.md for adding a screenshot

daminisatya:


daminisatya:
Now we need a screenshot for the apps listing page. So made it as an mandatory file and updated the apps.md. Since it is a small fix, Merging it.

Issue #331, opened 2016-05-10
start.sh always reports port 9000


treba123:
It doesn't read the config

shivenmian:
I think that's because the http port is specified as 9000 and the shortlink is localhost:9000.

What do you think it should be like otherwise?

treba123:
Well we have the option port.http=9000 in conf/config.properties.
I had set it to 80 and just wondered about the output
start.sh has the line
`echo "loklak server started at port 9000, open your browser at http://localhost:9000"`
hardcoded without caring what value port.http is set to. 

The shortlink has the port 9000 set again (shortlink.urlstub=http://localhost:9000) although i doubt it would work if it's not set accordingly to port.http

shivenmian:
Ah, I get it now. This is pertinent.

Orbiter:
when I put the 9000 in the start message I was too lazy to read the config. Furthermore, there is an option to overwrite the config in data/settings/customized_config.properties. That file must be parsed as well.

sudheesh001:
Assigning to @daminisatya 

This could be resolved by greping the line `port.http=` and reading the value then amending the shell file `start.sh`. 😄 Also check if the `data/settings/customized_config.properties` contains `port.http=` or `port.https=` if yes, then override the value.

PR #390, opened 2016-05-21
Added script to start.sh, to read the port value from the config file

daminisatya:


daminisatya:
selection_010

daminisatya:
selection_012

daminisatya:
#331

Issue #401, opened 2016-05-23
Deprecated Methods and outdated JRE


shivenmian:
There are some methods which aren't supported now and are outdated (the constructors are deprecated). The JRE version is also outdated.

Not a major issue, but will update them. @Orbiter

<img width="689" alt="screen shot 2016-05-23 at 2 20 39 pm" src="https://cloud.githubusercontent.com/assets/14081800/15465320/94e0e398-20f1-11e6-8da4-b81b2d7b6d5b.png">

shivenmian:
I realised @treba123 is working on the deprecated methods in his saveClientConnection branch, I updated the JRE in another PR.

PR #404, opened 2016-05-23
Fixed JRE version

shivenmian:
Was getting a warning about the JRE version being old (JavaSE 1.6) and build problems existing, updated it to JavaSE 8. #401


Orbiter:
We have not yet decided to go to java 8, but your code change does not force that for ant builds - thats ok

PR #412, opened 2016-05-24
Updated the Page's name on the nav bar

daminisatya:


daminisatya:
A small change. Merging it.

Issue #377, opened 2016-05-19
create basic peer information and create a peer hash


Orbiter:
right after the public key is created, check if there is a DAO.public_settings.peer_hash attribute.
If not existent, create it with this formula:
DAO.public_settings.peer_hash = base64(DAO.public_settings.public_key.hash())

This will make it possible that we can send a challenge to a peer to decrypt something we send them using the public key of that peer to encrypt something. Since the hash comes from the public key, only the creator of the key pair can be the owner of that hash. That makes it impossible to make a fraud hash.

It is important that a peer hash is only created when a key pair is created.

Orbiter:
this requires https://github.com/loklak/loklak_server/issues/375 first

treba123:
The standard hash function is not secure. I implemented it via sha256.
Also, direct acces to the values is not so cool in this case, so I created get/set function to take care of everything

treba123:
Hej @Orbiter regarding the hash thing: what you were refering to was [collision resistance](https://en.wikipedia.org/wiki/Collision_resistance) if i get you right.
So we want that it's extremly hard for an attacker to be able to create a key with the same hash as some other.
That's why we really shouldn't use anything weaker than sha256 for creating the hash, I'd actually prefer to use something like [bcrypt](https://en.wikipedia.org/wiki/Bcrypt) but didn't yet pull in any big library for it as sha256 is still regarded as pretty save.

If I get you right, you wanted to put constrains on the hash, like having so and so many zeros at the beginning. That's probably very unsecure, as it makes the room our hashes exist in smaller, making it actually easier to attack it (or did you want to make constrains on the key? anyway, both is actually dangerous)

treba123:
see above

Issue #375, opened 2016-05-19
create a secure peer key and a peer hash


Orbiter:
We need to ability to publish a public key and encrypt data from other peers with their public key.

Therefore each peer must create - if not already done - a public/private key pair.

To do:

- this uses the private and public settings from https://github.com/loklak/loklak_server/issues/374 do that first
- when the DAO class is initialized, look if a public key with name 'public_key' is stored in the public_settings. If not, create the key pair
- if you create the key pair, write the public key to public_settings.public_key, the encryption method as string into public_settings.public_key_method and the private key into private_settings.private_key

treba123:
As public_settings and private_settings are actually json-objects, shouldn't this be public_settings.get("public_key") instead of public_settings.public_key? Otherwise we duplicate data.

treba123:
Ok bullshit. We want to get key objects.

treba123:
Ok, to make the workflow more clear I added the function getPublicKey(), getPrivateKey() getKeyMethod() and getPeerHash() instead of direct acces of the variables.
Together with the corresponding set-functions

treba123:
see above

PR #418, opened 2016-05-24
Create or load keypair on startup

treba123:
This implements #375 and #377


treba123:
Ok this needs the move to java 8 first

PR #422, opened 2016-05-25
Fix NullPointerException in DumpImporter

shivenmian:
Need to test it more and provide a solution, right now I'm aborting startup if DAO.init() fails. Also, I also encounter a ZipException in DAO, when Zip file can't be opened.

Please don't merge, have to improve the commit.


shivenmian:
I'll submit an updated PR in a while, closing this.

Issue #413, opened 2016-05-24
IP Blacklist fails to deny access to the IP addresses listed


sudheesh001:

treba123:
Ah alright, I already anticipated this.
Will have a look into it

treba123:
@sudheesh001 I think it works now, could you do a test from this branch: https://github.com/loklak/loklak_server/tree/fixIpBlacklisting ?

treba123:
Fix works for me, please report again if it doesn't

PR #425, opened 2016-05-25
fix blacklisting

treba123:
This should fix the ip blacklisting https://github.com/loklak/loklak_server/issues/413

### DIFF
--- a/src/org/loklak/LoklakServer.java
+++ b/src/org/loklak/LoklakServer.java
@@ -466,16 +466,16 @@ public class LoklakServer {
         }
         
         // Setup IPAccessHandler for blacklists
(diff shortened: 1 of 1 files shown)


sudheesh001:
@treba123 Unable to ban the IP even now, I've added the IP : 192.168.8.104 to the 
```
server.blacklist=192.168.8.104
```
and browser to the server located at `192.168.8.102:9000` from the IP `192.168.8.104` and i am still able to access the server.

treba123:
Can you post you loklak.log (there shoud be some output about the blacklist)

treba123:
I tested this with 127.0.0.1 being blacklisted. That works fine.

Issue #419, opened 2016-05-24
Move to Java 8


treba123:
Before we can do this, we need to check if all one-click deployments support java 8

Branch: https://github.com/loklak/loklak_server/tree/java8
Please push changes needed for the switch in that branch instead of master, so we have don't have changes at different places.

Does support?
- [x] Heroku
- [x] Scalingo
- [x] Bluemix
- [x] Docker cloud

Checked if changes are needed?
- [x] Heroku (needs updated buildpack)
- [x] Scalingo (needs updated buildpack)
- [x] Bluemix (needs updated buildpack)
- [x] Docker cloud (the installation guide refers to ubuntu trusty instead of ubuntu:latest as the docker file. trusty does not support jdk-8)

Tested?
- [x] Heroku
- [ ] Scalingo (not sure yet, will see after merge)
- [x] Bluemix
- [x] Docker cloud

Other stuff:
- [x] Remove java 7 buildbot.

sudheesh001:
heroku, scalingo failing, needs updates on buildpack

treba123:
Ok, they still use openjdk 1.6. That's deprecated

treba123:
@sudheesh001 do you update the build packages?

sudheesh001:
Bluemix works. http://loklak-server-dispensary-1744.mybluemix.net/

Orbiter:
I just added https://github.com/loklak/loklak_server/commit/c6450cc4dc4b7ff10d83762cadb08d3e193a9e60
to support heroku and scalingo. Hints to update the remaining are included in the commit.

Please test heroku and scalingo deployment with this settings. Here is how you find out if the deployed loklak_server is actually running under Java 1.8:

- open /api/threaddump.txt on the server where you deployed your test Installation
- scroll down to the "ELASTICSEARCH ClUSTER STATS" section
- look for jvm.versions.version

Orbiter:
@sudheesh001 Bluemix does *NOT* work, see http://loklak-server-dispensary-1744.mybluemix.net/api/threaddump.txt according to my hint above: this is still running under Java 1.7

sudheesh001:
Yup just noticed, somehow it seems to run whereas the others (heroku/scalingo) crash

Orbiter:
We must have a confirmation from the thread dump, otherwise all peers will crash when we switch to Java 1.8

sudheesh001:
I've moved locally to java8, everything seems to be working but in the `ant` build i get a line as follows
[...]

Issue #378, opened 2016-05-19
every Timeline must contain the creators peer hash and public key


Orbiter:
extend the class https://github.com/loklak/loklak_server/blob/master/src/org/loklak/objects/Timeline.java 
and add two more metadata lines in Timeline.toJSON
- set peer_hash to DAO.public_settings.peer_hash
- set peer_public_key to DAO.public_settings.public_key

Orbiter:
do https://github.com/loklak/loklak_server/issues/377 first

treba123:
I'll also put key_method, as it is needed to use the key

treba123:
see above

treba123:
@Orbiter i don't exactly get why we should put the public key and the hash in the timeline. The hash can be easily calculated from the key, so it's just more to bits to carry. 
Wouldn't it make more sense to actually attach a signature for the content of the timeline and the hash? 
With the hash we can find the peer and ask him for his public key (which is pretty big) and the signature would make sure the content wasn't altered by third parties.
That would actually be like a email with signature

Orbiter:
Yes thats right, OK!

treba123:
Ok, i'll only write the hash for now and later do the signature

Issue #376, opened 2016-05-19
extend the '/api/hello.json' servlet - return the public settings


Orbiter:
change the return value of the 'hello' servlet:
- instead of returning the static 'ok' property, return the content of the DAO.public_settings JSON

Orbiter:
this requires https://github.com/loklak/loklak_server/issues/374 first

Orbiter:
@shivenmian please join in

treba123:
see above

PR #427, opened 2016-05-25
Secure peer key

treba123:
This implements #375, #377, #376 and #378 and is blocked by https://github.com/loklak/loklak_server/issues/419


treba123:
See https://github.com/loklak/loklak_server/pull/445

Issue #402, opened 2016-05-23
Null pointer Exception on running code


shivenmian:
When I run the code, I get a repeated NullPointerException. @Orbiter 

<img width="504" alt="screen shot 2016-05-22 at 4 59 42 pm" src="https://cloud.githubusercontent.com/assets/14081800/15465387/e4b10826-20f1-11e6-9e99-d4a90f799542.png">

treba123:
That looks like what i had with #332 . It just stopped for me though. Good to know that it happens for others, too.

treba123:
Ok, i just also encountered this. It happens for me when I had a not properly closed loklak still running, like ghosting around.
If that happens, port 9300 is still in use and i get the following error first:
`BindTransportException[Failed to bind to [9300]]; nested: ChannelException[Failed to bind to: localhost/127.0.0.1:9300]; nested: BindException[Address already in use];
	at org.elasticsearch.transport.netty.NettyTransport.bindToPort(NettyTransport.java:478)
`
If that happens, DAO aborts initialisation and therefor DAO.message_dump is null. I guess we have to abort startup if DAO.init() fails.

shivenmian:
Fixed in #432

PR #428, opened 2016-05-25
Fixed NPE in DumpImporter

shivenmian:
It happens when ```DAO.init()``` fails. This seems to be the only plausible solution right now, because the DAO can crash for a lot of reasons. It makes sense to abort loklak when DAO fails. Will concentrate on the individual bugs later. #402  

@Orbiter


shivenmian:
@treba123 gave a better fix, closing this.

PR #437, opened 2016-05-25
Used UI router to load charts as partials

### LINKED ISSUE #436: Include charts as partials in a single page
The user should be able to select a menu to see different charts instead of going to different pages, and also as many charts cannot fit on one screen at the same time

leonmak:
Fixes #436, user can select charts by choosing a tab, default set to stacked bar charts.


leonmak:
#453 solves this

Issue #459, opened 2016-05-27
Beautify the Force Directed Graph for the peers app


daminisatya:

daminisatya:
![selection_039](https://cloud.githubusercontent.com/assets/6227784/15595516/76671204-23dd-11e6-8814-a65a3ee3f76b.png)

Issue #462, opened 2016-05-27
Read the peers.json and convert it into a CSV file as required for forming FDG.


daminisatya:

daminisatya:
As of now the graph is reading the test json file which contains the peers.json object. We need to convert the .json file into CSV file and add the other required fields.

sudheesh001:
@daminisatya Is there a specific format you're looking for do you want all the fields in the current json to be in the csv ?

Orbiter:
why would you need a csv file, which software does that need?

sudheesh001:
@Orbiter To visualize the graph with d3, it's easiest to show a parent-child relationship with a csv as `d3.csv()` is built in and reads row by row making it easier, `d3.json()` too does the job but gets messy. I think the required format would be
`source(loklak.org),target(host),port.http,port.https,lastSeen,lastPath,PeerName`
Since this is only for the visualization to work, I think we can add `/peers.json?format=csv`

Orbiter:
Please implement ```peers.csv```. This can be done with the same servlet class. Please see https://github.com/loklak/loklak_server/blob/master/src/org/loklak/LoklakServer.java#L505-L507 and https://github.com/loklak/loklak_server/blob/master/src/org/loklak/LoklakServer.java#L525-L527 as example how the same servlet can return different contents.

sudheesh001:
Self assigning.

PR #463, opened 2016-05-27
#459- Beautified the graph

daminisatya:


daminisatya:
For a proper color differentiation, the graph required peers with different id's. This can be effectively viewed when the issue #462 is fixed.

PR #477, opened 2016-05-28
Fixes #391 Integrates the start of the advanced query generator into loklak

### LINKED ISSUE #391: Advanced query generator integration into loklak server
This is an enhancement. There are a lot of features that are present in the search, We have a basic tweet query search in the homepage itself but there are a lot of features that we have for the features like `from`, `until`,`since`, `count` etc.., Added to these we also have aggregations possible for hashtags etc.., Having these as checkboxes and the `since`, `until` as datepicker's we can have an advanced search page where the queries could be made as the user wants.

sudheesh001:
- [ ] Choosing the `hello.json` generates the required hello.json api query string
- [ ] Choosing the `peers.json` generates the required peers.json api query string
- [ ] Choosing the `status.json` generates the required status.json api query string
- [ ] The query string will take the base URL as the domain on which it's being run and not always redirect to loklak.org, when hosted on loklak.org it'll change to loklak.org
- [ ] The timezone offset value should always be automatically filled and the text box should not be editable.
- [ ] Clicking the `Query Loklak` button should query the api for the result and display the formatted `json` in the output textarea.
- [ ] Clicking on the `search.json` radio button should display the other fields needed to generate the `search.json` query string.
- [ ] Search.json queries should now be work dynamically on any change events.

Ensure that:

- [ ] Query parameter `q` is properly created
- [ ] Works across all hostnames and generically finds the hostname
- [ ] Additional `q` parameters like `since`, `until`, `near`, `from` should be constructed correctly and shown
- [ ] Additional GET parameters for `count`, `limit` and `minified` and `source` parameters in the URL, all of them should have the `&` making it a GET parameter of the query URL.
- [ ] Clicking the `Query Loklak` button should render the json response given by that URL


Orbiter:
please add also the new topmenu navigation

daminisatya:
@Orbiter added!

PR #480, opened 2016-05-29
Fixes #468 - Removed gaps and redesigned the apps page

### LINKED ISSUE #468: Fix Spaces and Gaps on Apps Page
The Apps page has a tile layout, with:
- gaps between the columns (thats good!)
- no gaps between the rows (they should be there and as wide as the gaps between colums)
- too much space on the left side and the right side:

apps

To fix this, do:
- add space between the rows
- show *four* columns instead of *three*
- make the gap on the left and right side of the tile matrix as big as the gap bewteen the columns
- keep the ability to re-size the tiles as the window width is changed.

daminisatya:


daminisatya:
selection_045

daminisatya:
selection_046

PR #495, opened 2016-05-31
Added base64 Encoding

shivenmian:
Encoding passwords in Base64 to avoid problems with unicode chars.


shivenmian:
This is a small addition so merging it.

Issue #510, opened 2016-06-02
Clean up of json files


treba123:
The json-files we use for the AAA system currently never get cleaned up, making them grow forever and saving user data (like ips) in clear text.

As I need some kind of cleaup system for the DAO.authentication file anyways, I'd like to implement a cleanup function (has to be thread-save) that removes all entries with a `valid_until` key that holds a timestamp smaller then the current one.
It has to be triggered by a thread that wakes up maybe every ten minutes or so, triggering the cleanup for the corresponding files.

Edit: to be future save, we should use long as the timestamp format.

Orbiter:
We should not store data in the AAA files which then have to be cleaned up. This effect is circumvented with https://github.com/loklak/loklak_server/commit/dc5712152e0fe293abc132529672c54ece6dbb0e where volatile information is never stored in the json files.

PR #518, opened 2016-06-03
Cleanup thread for AAA files

treba123:
This adds a cleanup thread for the AAA files, so we don't carry outdated information for ever, see https://github.com/loklak/loklak_server/issues/510

Right now it removes the following entries from the authentication file:
- anonymous logins (after 24h of inactivity)
- cookies (after a week of inactivity, which is the current expire time for long living cookies)
- it does not remove any entries without `expire_on` entry

The thread is modeled like the other threads to be in line and make it easy to understand, although technically, it probably should have been done as task with a tasksscheduler.

@Orbiter you removed the expire entries on the authorization file in https://github.com/loklak/loklak_server/commit/96fd4e7e186822756b96f0c2abc21c22ab96615f , but I'd suggest to clean it up aswell, as for every anonymous user, an authorization entry with an ip is made and carried around forever.
If you don't plan to change that anyway, we need to remove them after some time.
Same goes for accounting, if we start to make it saved permanently.

### DIFF
--- a/src/org/loklak/LoklakServer.java
+++ b/src/org/loklak/LoklakServer.java
@@ -103,6 +103,8 @@ import org.loklak.http.RemoteAccess;
 import org.loklak.server.FileHandler;
 import org.loklak.tools.Browser;
 import org.loklak.tools.OS;
+import org.loklak.tools.storage.DatabaseCleanupThread;
+import org.loklak.tools.storage.JsonFileAAA;
 
 
 public class LoklakServer {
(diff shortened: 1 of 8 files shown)


Orbiter:
The AAA storage files are supposed to be written rarely. To set an expire date each time when the server accesses the Authentication database creates IO every time. That should not happen and it is not required at all. Events that are to be measured must go to the Accounting object. We should have two account databases: one only in RAM and one to be stored.

Orbiter:
This solution is not efficient. As a replacement, I added https://github.com/loklak/loklak_server/commit/dc5712152e0fe293abc132529672c54ece6dbb0e which either stores credentials or identities in persistent or volatile way. This removes the requirement for a clean-up.

treba123:
That's sound. I just hope it won't become a memory leak on long living instances (as the information is still carried around, even only in ram).
On the long run, we should consider using a real database that can handle this problems for us.

PR #522, opened 2016-06-04
Added Password Recovery page

shivenmian:
Added the page to be shown when "Forgot Password?" is clicked on the Login page. This is a small webpage which works, the more important part is the Email functionality that I'm currently writing, so I'm merging this.


html/apps/forgotpassword/js/passwordrecovery.js
```diff
@@ -0,0 +1,32 @@
...
+    var emailerr = false, session = true;
+
+    $('#submit').click(function(){
+        checkEmpty();
+        if(emailerr){
+            alert("Please fill email");
+        } else{
+            var mail = encodeURIComponent($('#email').val());
+            alert("Email: " + mail + "\nSuccess, write post function");
+            //post function
```
sudheesh001: Why is thi missing ?

html/apps/forgotpassword/js/passwordrecovery.js
```diff
@@ -0,0 +1,32 @@
...
+{
+    var emailerr = false, session = true;
+
+    $('#submit').click(function(){
+        checkEmpty();
+        if(emailerr){
+            alert("Please fill email");
+        } else{
+            var mail = encodeURIComponent($('#email').val());
+            alert("Email: " + mail + "\nSuccess, write post function");
```
sudheesh001: Please move to console log's and not alert's, too many alert's results in browsers complaining about suppressing the popups. We've spoken about this before

html/apps/forgotpassword/js/passwordrecovery.js
shivenmian: Because the function hasn't been added. I will add the PHPList lines to ```//post function```.

Issue #467, opened 2016-05-27
Add Email Server for Password Recovery


shivenmian:
We need three things:

- [x] On clicking of "Forgot password", lead user to a page where he enters his email ID for password recovery in the Login page (see #443). Create this page.

- [x] A server which sends an email to that email ID with a recovery link (autogenerated). The link has to have a hash, url will be something like ```/passwordreset/<hash>```.

- [x] That link leads to a "Set new password" page. Create this page and the aforementioned server.

I will use the class made in #424 for the server and create the relevant webpages as well.

shivenmian:
Anything more you'd want me to add @Orbiter?

shivenmian:
For the email server, I came across two options as of now:

* [Spring](http://docs.spring.io/spring-integration/reference/html/mail.html)
* [JavaMail](http://www.oracle.com/technetwork/java/javamail/index.html)

Both have clear docs and seem to be secure, and can be implemented into the Java code well.

For password recovery, what I think should be done is this:

1. In the "Forgot Password" page, user enters his email ID (which we will check from database if it exists). 

2. If it exists, we will create an autogenerated recovery link (something like ```/passwordreset/<hash>``` and mail it to the user on that ID. The recovery link leads to a reset password page (which I will create), wherein user chooses the new password and it is written onto the database.

@Orbiter What email server do you prefer? Is the idea fine? And should I get started on this?

Orbiter:
JavaMail would also be my choice!
The process you describe is also what I would believe could be the correct way.

shivenmian:
@Orbiter nice. So should I start on this?

Orbiter:
yes!

smokingwheels:
For what its worth my Provider Blocks Port 25 so would it be possible to use Yahoo, Gmail or other as the SMTP Server you need to put a custom port in and user name as Email address and matching password.

shivenmian:
This could take me a day or two.

mariobehling:
Ok, we should not take care of this. Please check how to make use of phplist.com. The CEO is a friend of mine and they will offer a large free plan for loklak.

mariobehling:
@shivenmian Did you have a chance to check phplist out? How much does it help us achieving this goal? What is your opinion? Should we head this way and integrate?

shivenmian:
@mariobehling I'll let you know in a bit, probably by this evening.

shivenmian:
[...]

PR #553, opened 2016-06-09
Added basic email structure

shivenmian:
@treba123 @mariobehling @Orbiter I've added a basic email structure for the Password Recovery system in #467.

Since we do not have a server set up as of yet (we need one), I set the sender email as server@loklak.org and password as randomxyz (its a random name). We can change it later.

The user gets to set up his SMTP type, his hostname and port etc. in config file. Please see the options I have added there. He also has to make changes on the server side (for example if host is GMail then POP3 and SMTP should be turned on in their Settings, else mail will be found as suspicious). The email will then be sent.

I have kept the email body empty, because I just wanted to see if this structure is right. Once this is approved, I will simply make the Password Recovery HTML page (to set a new password) and put its link (I will append a hash to the link) in the body and the process will be complete.

I have not tested this because I don't have a server myself. But the build is successful though. Please review.

### DIFF
--- a/conf/config.properties
+++ b/conf/config.properties
@@ -7,6 +7,22 @@
 # allow public signup
 users.public.signup=false
 
+# SMTP emailing switch (for password recovery and signup verification etc)
+smtp.mails.enabled=false
+
+# SMTP password recovery server host (change this as per need)
+smtp.host.name=smtp.gmail.com
+
+# sender email (change this as per need)
+smtp.host.senderid=server@loklak.org
+
+# depending on security of recovery host server, set authtype as none, starttls or tls
+# also make changes on server side: eg if server is gmail, turn POP3 and SMTP on
+smtp.host.encryption=tls
+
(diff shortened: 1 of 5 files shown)


treba123:
Two notes:
- there's the LoklakEmailHandler dummy class already, maybe it's nicer to move the pure email logic there so we can use it everywhere, as there's many other usescases for email. The servlet should only contain what's needed to send the recovery email and reset the password
- in the config, maybe make the key names more informative, like `smtp.host.name` instead of `host.name`

treba123:
Today I'll send a pr that will make the setting of a new password more easy/clear and add the email verification. Maybe you want to have a look on that for how I create a token for the email

shivenmian:
@treba123 can you check now? If any more optimisation needed, please tell me.

shivenmian:
I have one small optimisation left, gimme 2 mins.

PR #556, opened 2016-06-09
Email verifications for signups and lots of refinement

treba123:
The emails do not get send yet, as the class for that is not yet done.
But a link is being created and work.

The design here is to create a login token (in this case only usable for one time) that logs the user in, and a request to activate his login via email/password.

@Orbiter I made some little changes to the Authentication and Authorization classes, hope that get's in line with you. I wanted to avoid code duplication and therefor moved some code in the classes that were otherwise needed outside.


shivenmian:
Please check my PR before you merge.

PR #563, opened 2016-06-10
add checks in signup servlet, make password configurable via config, …

treba123:
…let signup app get that parameters via ajax


shivenmian:
Works :)

Issue #566, opened 2016-06-10
Add charts for visualizing the result by Sentiment Visualizer app. 


daminisatya:
Add Pie Charts for visualizing the following

- [ ] Countries tweeting the tweets.

- [ ] Languages of the tweets

- [x] The percentage of emotion tags being categorized.

PR #588, opened 2016-06-14
Fixes: #566 added pie chart for viewing the tweets languages

daminisatya:


sudheesh001:
Clean patch ! Good work @daminisatya

PR #595, opened 2016-06-15
Cookie support to JSON objects in Authentication objects

sudheesh001:
This is useful once we have the Authentication in place and when the IOT devices will be sending over the authentication tokens to loklak server, they can use this to convert the Cookies to JSON so that the connections can be maintained.

### DIFF
--- a/src/org/json/Cookie.java
+++ b/src/org/json/Cookie.java
@@ -1,0 +1,169 @@
+package org.json;
+
+/*
+Copyright (c) 2002 JSON.org
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+The Software shall be used for Good, not Evil.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


sudheesh001:
This is a part of the org.json , could be used potentially for managing the cookies.

treba123:
Hm do you really think that cookies are the right things to use here? Arent cookies more something for the browser world?
Also, how would they be created?
Other APIs often just suppert a get-token API, where you once log-in with you username and password and recieve a token (the cookie also just contains such a token).
Wouldn't that be easier? 

I can add `getToken` servlet if you want, which can be used by the IoT devices.
The flow would then go as follows:
- the device calls 'server.com/api/getToken.json' which returns a token
- on each following call, the device just adds `loginToken=...` to the query
- if the token is not used for too long, it gets removed and the device has to login again (i'd call that handshake)

sudheesh001:
@treba123 precisely thats exactly what we need, tokens with secret key and consumer key so that the application/IOT device can communicate with the server. These files are just string conversions and helper functions so that you could implement the getToken, The device when sends the tokens (can be treated as a cookie) and converted using the functions like `toString()` or from String to JSON Object, so these json objects have the required KV pairs needed for doing the authentication.

treba123:
@sudheesh001 sorry, I still don't really understand you.
Why do we need cookies here?

We already have token login, creating a cookie around it would create additional overhead, and for the 'getToken' servlet we dont't need it (in the end, a cookie is just a more complex entity which holds some information, in our case some kind of login token).

Overall, cookies are usually not something you want to use in a restful/stateles api. You only use request headers and parameters.

Cookies are something you use in a stateful communication, for example in browser sessions.

Sorry, I appreciate your work here, I just still don't see the usecase. Can you help me?

treba123:
I created a servlet, found in api/handshake-client.json, which gives you an access token.
You can specify `valid_seconds=600`, the time how long it is valid. `valid_seconds=-1` makes it valid forever.
The usecase is now the following:
- call `domain.com/api/handshake.json?login=bla@blubb.de&password=123456&valid_seconds=3600` and in return you get a access_token (for example YYqs62uSqSYfZwufdZN1rPLbqGTyaA)
- to all further requests, add `access_token=YYqs62uSqSYfZwufdZN1rPLbqGTyaA`

PR #606, opened 2016-06-16
Use the APIServiceLevel to limit access to servlets

treba123:
This makes APIServiceLevel comparable and makes AbstractAPIHandler return a 401 if the service level a user has does not match the default service level.

I created that as part to understand how APIServiceLevel is meant to work, for discussion.

The idea behind it is that, for propper API, if a user does not have the rights to do something, we should respond with a propper http status.
Currently, this is only possible within AbstactAPIHandler, not in a servlet extending it (actually we can return null, but then a 400 is responded, which is not always the right answer).
Therefor it would be desireble if AbstactAPIHandler checks if the user has the basic rights to access a servlet.

### DIFF
--- a/src/org/loklak/api/cms/LoginService.java
+++ b/src/org/loklak/api/cms/LoginService.java
@@ -39,7 +39,7 @@ public class LoginService extends AbstractAPIHandler implements APIHandler {
 
     @Override
     public APIServiceLevel getCustomServiceLevel(Authorization rights) {
-        return APIServiceLevel.ADMIN;
+    	return APIServiceLevel.PUBLIC;
     }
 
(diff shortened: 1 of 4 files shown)


Orbiter:
Well, the semantics of the change is good and correct while the usage of the Comparable property of an Enum is slightly bend - even without using a explicit number to order an enum object the level number of an ServiceLevel would be available with Enum.ordinal(). We would have to change the list order of the APIServiceLevel (i.e. into PUBLIC, PRODUCT, LIMITED, ADMIN) to just be able to use '>' and '<' to compare the level instead of the newly introduced comparing methods.

PR #594, opened 2016-06-15
Added Initial scraper which scrapes all the links from a given URL

daminisatya:

PR #608, opened 2016-06-16
Added Sitemap and robots.txt + small fixes

### LINKED ISSUE #465: add a robots.txt and sitemap.xml
Because our site navigation cannot be crawled so easily with a non-javascript enabled web crawler since https://github.com/loklak/loklak_server/issues/394, we need a sitemap which links all available pages.

- implement a sitemap servlet in org.loklak.api.cms.Sitemap.java which returns xml according to https://en.wikipedia.org/wiki/Sitemaps
- the sitemap shall list all pages using the method org.loklak.api.cms.TopMenu.service(); - like (new TopMenu()).service(). Then you get all paths as TopMenu presents that as well.
- add a robots.txt file in https://github.com/loklak/loklak_server/tree/master/html as static file
- the robots.txt must allow all search engines to load all pages, see https://en.wikipedia.org/wiki/Robots_exclusion_standard
- the robots.txt must include a link to the sitemap according to https://en.wikipedia.org/wiki/Sitemaps#Search_engine_submission
- for security, add a crawl-delay of 10 seconds to robots.txt

shivenmian:
Fixes #465. 

PS: I earlier kept baseurl as the base url of the sitemap.xml (this can change depending on localhost, loklak.org and oneclick deployment links). But since I had to add to robots.txt, I kept it as loklak.org for now.

This sitemap will keep updating itself with the topmenu service, so any time any link is added to loklak, sitemap updates.

Also added JSoup to classpath (opening the project in eclipse was giving me errors because it wasn't in the classpath), fixed the links in TopMenu servlet, and added a missing serial ID. These are small fixes.

@Orbiter @mariobehling please check and tell me for any modifications.

### DIFF
--- a/src/org/loklak/api/search/GenericScraper.java
+++ b/src/org/loklak/api/search/GenericScraper.java
@@ -48,6 +48,8 @@ import org.jsoup.select.Elements;
 
 public class GenericScraper extends HttpServlet {
 
+	private static final long serialVersionUID = 4653635987712691127L;
+
 	@Override
 	protected void doPost(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
 		doGet(request, response);
--- a/.classpath
+++ b/.classpath
@@ -68,5 +68,6 @@
 	<classpathentry kind="lib" path="lib/t-digest-3.0.jar"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
 	<classpathentry kind="lib" path="conf/logs"/>
+	<classpathentry kind="lib" path="lib/jsoup-1.9.2.jar"/>
 	<classpathentry kind="output" path="classes"/>
 </classpath>
--- a/html/robots.txt
+++ b/html/robots.txt
@@ -1,3 +1,2 @@
 User-agent: *
-Disallow: /api/
-Disallow: /dump/
+Sitemap: http://loklak.org/api/sitemap.xml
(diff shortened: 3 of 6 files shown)


Orbiter:
ok, looks good but
- why do you need jsoup?
- why do you remove the disallow entries from the robots.txt?

shivenmian:
@Orbiter I didn't need JSoup. I opened Eclipse after pulling the changes from #594 and it was showing me errors because JSoup was not in the classpath, so added that. It isn't related to this issue.

I removed those lines because you wanted the search engines to crawl all the pages as mentioned in the issue.

src/org/loklak/api/search/GenericScraper.java
```diff
@@ -48,6 +48,8 @@
 
 public class GenericScraper extends HttpServlet {
 
+	private static final long serialVersionUID = 4653635987712691127L;
```
sudheesh001: This isn't related is it ?

shivenmian: It is autogenerated and not having it gives a warning. @daminisatya hadn't added it in her PR, so I did that.

shivenmian: @sudheesh001 edited title of the commit + description acc to this. If the rest looks ok to you, can this be merged? @Orbiter

sudheesh001: @shivenmian I think you can keep the disallow for dump and api

daminisatya: @sudheesh001 @shivenmian If this particular file patch is creating a problem. I can fix it right now.

shivenmian: @sudheesh001 The issue suggests that all the URLs should be crawled, so why keep them?

PR #614, opened 2016-06-21
Added reset servlet + password change functionality

shivenmian:
This is a fix for #467 Part 3. I'm unsure if this will work because I haven't tested it so I do need feedback. 

This code first sends email with a reset link once password change is requested (reset link has a token). This reset link is sent as a param to the resetpass app (yet to code), and through the token, we can display to the user which email ID's password is being changed. I have decided to keep a file named ```passwordreset.json``` which stores tokens as keys, and the value of every token is a JSONObject which again consists of id, one_time and expiry time. Mapping keys to ID (rather than the other way round) was more convenient for me to reduce code complexity.

Once password is changed, I will check on the client-side (i.e on app) itself whether password is right etc, and then token and new password is posted to ```resetpassword.json```, and through token we figure out the email ID in ```DAO.authentication```, and set the new password.

I will code up the resetpass webpage (as described above) ASAP, it's not a big deal now that I've coded this.

Please review and give me feedback in this if there is anything wrong, I will fix it. @Orbiter @treba123 @mariobehling.

### DIFF
--- a/conf/templates/reset-mail.txt
+++ b/conf/templates/reset-mail.txt
@@ -1,0 +1,7 @@
+You just requested for a password reset for your loklak account.
+
+To do so, click on the following link:
+
+%RESET-LINK%
+
+If you didn't ask for a password reset, please ignore this email.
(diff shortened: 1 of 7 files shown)


shivenmian:
This is review+. Since @treba123 has sent PR #616 which may add conflicts, merging this. Will add Password Reset page + regex check in server side in another PR.

PR #623, opened 2016-06-22
Added Reset Password Page + minor formatting

shivenmian:
@mariobehling @Orbiter @treba123 Added the Password Reset page. Also did some formatting of the SignUpService (Source --> Format) so that the indentation is right, you can ignore that.

I tried with queries like ```localhost:9000/apps/resetpass/index.html``` and ```localhost:9000/apps/resetpass/index.html?token=1234``` and I got "Specify token" and "Invalid token" respectively. 

Please test it and review.


html/apps/resetpass/js/resetpass.js
```diff
@@ -0,0 +1,183 @@
+$(document).ready(function()
+{
+  var passerr = false, confirmerr = false;
+
+	// get password parameters
+	var regex;
+    var urltoken = getParameter('token');
```
daminisatya: Please check your indentation.

shivenmian: done

PR #626, opened 2016-06-23
Always respond with an http status code if a request was not successf…

treba123:
…ul (in signup)


shivenmian:
@treba123 looks fine to me. Resolve conflicts, rest is ok.

PR #630, opened 2016-06-23
make travis only build once on PRs from branches

treba123:
At the moment, the travis build builds the project two times when doing a PR to master from a branch of the loklak repo. This should fix that

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,7 @@ script:
   - bin/start.sh
 
 install: true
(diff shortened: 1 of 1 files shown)


treba123:
found here: https://github.com/travis-ci/travis-ci/issues/1147

sudheesh001:
@treba123 I don't think this should be done, Why not test other branches whenever a push is made ?

PR #641, opened 2016-06-29
Eventbrite Crawler: JSON Object modifcations wrt Open Event format

jig08:
JSON Files written out wrt #640 in `<user>/Downloads` folder as `event.json` , `org.json` and `social_links.json`


shivenmian:
Please squash your commits always.

Issue #622, opened 2016-06-22
Creating a twitter analysis app


shivenmian:
In one of my college's research groups, a friend had made [this project](http://twitdigest.iiitd.edu.in/fbanalysis). It makes you login through FB and then gives you statistics like your FB activity (what % of time you updated status, put up photo etc), an activity chart of FB activity, likes statistics (total likes, who likes your activities the most), word cloud of most popular words and sentiment analysis etc. (A sample analysis can be seen [here](http://twitdigest.iiitd.edu.in/fbanalysis/sampleprofile/)).

We could implement a similar app for Twitter data as well. We also get an added benefit by using loklak: the user will not have to log onto twitter at all. He can simply add the username of any user and he gets the data, so this is definitely more useful. 

As a **first step**, I'll create a webpage wherein I specify the username I need the analysis for, and clicking the submit button returns a JSON API with all the data. The JSON will obviously be implemented via a servlet, so **creation of the servlet is the main part**. Once the servlet creation is done, I'll come onto using the JSON data and making the entire frontend and UI.

Looking at the app I showed above (look at the sample analysis), I propose the following features to be implemented into the servlet. Please suggest more if you can:

- [x] Activity frequency: tweet frequency graph (yearly, hourly and daily).

- [x] Activity pattern: what number of times user uploaded photo / video  etc

- [x] Activity likes and retweets: Number of likes you did, number of retweets, the page / user's posts you like the most, top 5 people who like / retweet my posts the most, likes frequency.

- [x] Content stats: sentiment analysis chart of content, words / hashtags I use the most

I had added this idea in the Questions doc Mario made. Please tell me if you like the idea and give suggestions / features to be implemented. I can then start working on it. @mariobehling @Orbiter

shivenmian:
@Orbiter basically for implementing the app, all I need is the data from my own tweets, data of followers I have, and data of following.

I had another look at all the loklak APIs, and I believe four of them: ```search.json```, ```account.json```, ```user.json``` and ```asset``` will give me all the required information about the user. So I will be using these four. 

Or do you think I may have missed something?

PR #665, opened 2016-07-07
Added Twitter Analysis app servlet

shivenmian:
Working on Issue #622. Will be adding code for every part in separate commits, merge at the end.


src/org/loklak/LoklakServer.java
```diff
@@ -651,4 +652,4 @@ private static void checkServerPorts(int httpPort, int httpsPort) throws IOExcep
 	        }
         }
     }
-}
+}
```
daminisatya: @shivenmian PLease leave a line after 655.

shivenmian:
There are some more bugs I have found out in the search servlet, I will fix them too and work on the last 2 parts of #622. Merging this for the time being.

PR #669, opened 2016-07-08
Added initial set of synonyms to susi_cognition_000.json

daminisatya:


daminisatya:
@Orbiter, please review this initial set, If this is what being expected i will go ahead and add the rest.

Orbiter:
Hi Damini, that is a good list but they are not synonyms but categories. Synonyms are words that you can literally replace and do not change the semantics of a phrase. Categories are words that describe which other words have a close semantic connection, like family:{mother, father, brother, sister}.

However I merge your request and move the words to the catories later

Issue #677, opened 2016-07-12
StringIndexOutOfBoundsException in eventbrite parser


Orbiter:
There was an example url in the pull request https://github.com/loklak/loklak_server/pull/676 which showed an error in the eventbrite parser.

try ```http://localhost:9000/api/console.json?q=SELECT%20*%20FROM%20eventbrite%20WHERE%20url=%27https://www.eventbrite.com/e/?q=global-health-security-focus-africa-tickets-25740798421%27%27;```

causes
```
java.lang.StringIndexOutOfBoundsException: String index out of range: 19
	at java.lang.String.substring(String.java:1951)
	at org.loklak.api.search.EventBriteCrawlerService.crawlEventBrite(EventBriteCrawlerService.java:118)
	at org.loklak.api.search.ConsoleService.lambda$11(ConsoleService.java:330)
	at org.loklak.api.search.ConsoleService$$Lambda$15/746247411.apply(Unknown Source)
	at org.loklak.api.search.ConsoleService.console(ConsoleService.java:346)
	at org.loklak.api.search.ConsoleService.serviceImpl(ConsoleService.java:363)
	at org.loklak.server.AbstractAPIHandler.process(AbstractAPIHandler.java:162)
	at org.loklak.server.AbstractAPIHandler.doGet(AbstractAPIHandler.java:109)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:687)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:845)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:583)
	at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1174)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:143)
	at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:134)
	at org.eclipse.jetty.rewrite.handler.RewriteHandler.handle(RewriteHandler.java:335)
	at org.eclipse.jetty.server.handler.HandlerList.handle(HandlerList.java:52)
	at org.eclipse.jetty.server.handler.gzip.GzipHandler.handle(GzipHandler.java:418)
	at org.eclipse.jetty.server.session.SessionHandler.doHandle(SessionHandler.java:226)
[...]

PR #680, opened 2016-07-12
Fixing NPEs and other StringIndexOutOfBoundsException in eventbrite parser

jig08:
… wrt #677

screenshot from 2016-07-12 17 10 48


daminisatya:
@jig08 was this tested properly?

jig08:
@daminisatya Well, I did try covering all the places which could yield `NULL` and added a check for it. Do you still face any `Null Pointer Exception` ?

PR #683, opened 2016-07-13
Finished Twitter Analysis app

shivenmian:
Finishing the last two parts of #622, will be adding a few commits, do not merge.


shivenmian:
The entire app is done, except one thing. I wanted to code up a function which returns the most frequently meaningful used words, and that'll use Lucene's Keyword Extraction. Will code it a bit later once I get into Susi's NLP code. Rest all done.

mariobehling:
@shivenmian Not sure why you are merging your own commit in no time. Commits should be merged by other people on the team and tested before that. Where is the app? Screenshots? Tests?

PR #685, opened 2016-07-13
Removing spaces in JSONArray key in EventBrite Parser

jig08:


jig08:
Minor Change, hence merging this.

PR #710, opened 2016-07-17
Login changes, key registration and login, lots of small changes

treba123:
This is a rather big PR with lots of login related code but also some other small fixes I made on the way.

The main features are public key registration/creation, public key login (roughly modeled after ssh2).

smaller fixes include:
- httpservletqueries are now explicitly in UTF-8 format (instead of ISO_8859_1). This solves some encoding issues we had in some app before
- the log system is now closed at the end, not together with other threads
- all login code related to passwords is now moved to the login service
- client handshake removed. use login service instead
- some documentation
- many more small fixes


shivenmian:
+1

PR #714, opened 2016-07-18
Propper key conversion from DER to PEM format (as used with openssl)

treba123:
With this commit, key can in PEM format can now be registered and are also given when a new key-pair is created by the server.
That should make it easier for IoT devices to use the pubkey login, as the PEM format is much wider used and supported (for example by openssl), especially in non-java projects.


mariobehling:
@treba123 Which issue does this PR belong to? Please help us and make it easier to follow up on the project development.

Please add the relevant issue to each PR you make.

Please do not merge your own PRs without peer-review.

PR #718, opened 2016-07-19
Added support for the dutch language

cribsb:
I added the top 100 dutch words to the configuration file and I added the Dutch language to src/org/loklak/data/Classifier.java


Orbiter:
wow! great thing, thank you!

Issue #668, opened 2016-07-08
Add initial set of categories to the susi_cognition_000.json


daminisatya:

daminisatya:
#722

PR #722, opened 2016-07-20
Added some more categories to susi_abstraction_en JSON file

daminisatya:
Categories , fillers and synonyms are three terms which are being used to classify an expression from the Susi rule. Susi reads a particular expression and categorizes based on the key terms and understands the meaning of it. This PR had some general words(which are categories) each with a set of other related words (related to the meaning of the root word). This is the initial category list added to Susi.


mariobehling:
@daminisatya Description and link to the relevant issue is missing. Which issue does that PR belong to?

daminisatya:
@mariobehling #668

mariobehling:
Description is missing. What categories and how you sort them? Either provide info in the issue or in the PR.

Orbiter:
great, thank you!

PR #724, opened 2016-07-20
HTTPS portability

treba123:
This commit fixes a portability issue to platforms without openssl or "SUN" security provider.
It also simplifies the key conversion code.

Edit: the new library will probably also help on future work in other areas. For example support of curve25519-signatures, which are much more performant (faster and smaller key-sizes) than RSA. This is mostly interesting for IoT or embedded devices (where this is critical), but also could benefit peer-to-peer workloads.


mariobehling:
@treba123 Which issue is it related to?

treba123:
@mariobehling Didn't open an issue before, just faced the problem myself and fixed it directly. Sorry if that is a problem.

mariobehling:
Thanks for the info. We had a number of cases where people worked on overlapping issues now - without creating an issue.

It might not be the case in your area, but it would make it much easier if we all follow best practices.

So, let's match every PR with an issue and provide some short description (which would make it hopefully easier and faster to review as well).

Thank you!

mariobehling:
@treba123 Please ask someone to merge your PRs. I did not merge, because I am not into this topic. I think @shivenmian is involved usually. You could have asked him. We should follow peer-review practices.

PR #735, opened 2016-07-21
Add helper function for parsing Nested Html and performing request on Weibo

DengYiping:


daminisatya:
@DengYiping Please give me a sample screen shot of the result or an example so that I can test it once :)

DengYiping:
Please read 
[https://github.com/DengYiping/WeiboSpider/blob/master/src/test/resources/test.html](url)
as a String in Java program and call the helper function.
You will see something like:
[https://github.com/DengYiping/WeiboSpider/blob/master/src/test/resources/test1.html](url)

daminisatya:
@DengYiping For some reason the URLs are broken. Giving 404.

DengYiping:
@daminisatya Fixed

daminisatya:
@DengYiping How can we get the source code from weibo pages. Looks like I'm unable to get it from the URL.

DengYiping:
@daminisatya You have to login to get the source code. See here: 

[https://github.com/DengYiping/WeiboSpider/blob/master/src/main/scala/com/sd/loader/UserInfoLoader.scala#L16](url)

jig08:
@daminisatya @DengYiping  Scraping nested links in HTML using JSoup shall be helpful for other scrapers too. Let know how to use it once it is completed.

DengYiping:
@daminisatya This helper is finished. Weibo store all its html contents in script, that's why I create this helper function.

DengYiping:
@jig08 The helper is finished. I just wait for it to be finished. Further exploration on weibo is under taking. @mariobehling Please merge this branch.

daminisatya:
Let me review the code once. I would like to test it on my code.

jig08:
@DengYiping Yeah Sure.

daminisatya:
**Question :** @DengYiping Is there any other way other than using cookies? Anyone who want to use the service they have to update their browser cookie and update the code!

DengYiping:
@daminisatya I've implement a cookie pool. If we manage a cookie pool well enough, it won't be a problem. It is the best I can get.

daminisatya:
@DengYiping Awesome. I shall integrate this to the scraper PR. Please help me do that :+1:

DengYiping:
@daminisatya OK, how we should do that?

src/org/loklak/harvester/WeiboInfoScraper.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.regex.Matcher;
+
+public class WeiboInfoScraper {
+	/**
+	 * This is a helper function that helps user to extract html nested inside of html script
+	 * @param raw_html
+	 * @return nested html String
+	 */
+	public static String getNestedHtml(String raw_html){
+		String html = raw_html.replace("\\","");
```
daminisatya: Can we pass the URL directly to ```getNestedHtml``` , call the ```buildRequest``` and get the html ? So that i can just use getNestedHtml method directly. 
Also what folder structure do you suggest? Can we create a weibo folder and add all the request files into it?

DengYiping:
@daminisatya , OK, I will implement that.

DengYiping:
@daminisatya I think that will bypass inserting cookies. I will implement another wrapper on top of two to handle the problem.

DengYiping:
@daminisatya There you go. By the way, what does this pink "SUSI" tag means?

DengYiping:
@daminisatya Wait a minute, minor changes.

daminisatya:
@DengYiping The tag makes easy for categorizing the PR. Susi is a personal assistant which is currently under development stage. Susi answer queries related to the services available in Loklak. So this weibo scraper (if completed) can be integrated into Susi's intelligence! Try this asksusi.com

DengYiping:
@daminisatya Just fix a small problem, ready to go.

jig08:
@DengYiping Could you kindly include a sample as to how nested links can be scraped using this ?

DengYiping:
@jig08 This one is not for nested link. This helper function is for crawling weibo. It will decode the html nested inside javascript(weibo do it to anti-crawler.)

jig08:
@DengYiping Oh great, I was looking for a function like this for the Instagram crawler. Any insights on how it can be reused with other scrapers ?

DengYiping:
Minor modification might be needed. You could try to pass a raw html string into the parser.

jig08:
Okay, thanks. I'll give it a try.

Issue #738, opened 2016-07-21
Include rules for Date-Time in SUSI 


jig08:
Using the console service implemented in PR https://github.com/loklak/loklak_server/pull/719, expand the ability of SUSI.

daminisatya:
@jig08 Will this date-time service give answers for questions like

- Difference between timezones?

- What is the current time now?

- Timezone of a particular place?

- Date of a particular place? 

- Time of a particular place?

Im curious on what kinda rules this service can provide?

jig08:
@daminisatya Currently, it aims to answer queries like:
* What is the time ?
* What is the date today ?
* Which day of the week is it ?
* What timezone are we in ?
etc.

For questions like:
>Difference between timezones?
What is the current time now?
Timezone of a particular place?
Date of a particular place?
Time of a particular place?

I would have to make use of a third party service like [www.timeanddate.com](http://www.timeanddate.com/) or any other

sudheesh001:
> What is the time ?
What is the date today ?
Which day of the week is it ?
What timezone are we in ? etc.

For these questions we can directly query from the client itself in susi rather than request and load the server, the aim here should be to scrape timeanddate.com and display the contents after the required computation for the query given. Please refer https://github.com/loklak/loklak_server/pull/745#issuecomment-234462082

sudheesh001:
Moving to #748

Issue #732, opened 2016-07-20
Adding Date-Time service in console for SUSI


jig08:
Similar to other services like `wordpress`, `meetups`, `eventbrite` etc, creating a console utility which enables _SUSI_ to answer queries related to date and time

sudheesh001:
Moved to #748

PR #745, opened 2016-07-22
Splitting date & time for useful info

jig08:
In order to answer queries (wrt #738 )
Follow up of issue https://github.com/loklak/loklak_server/issues/732 (Adding Date-Time service in console for SUSI)

_EDIT:_  New issue corresponding to this https://github.com/loklak/loklak_server/issues/747

### DIFF
--- a/src/org/loklak/api/search/TimeAndDateService.java
+++ b/src/org/loklak/api/search/TimeAndDateService.java
@@ -62,7 +62,30 @@ public abstract class TimeAndDateService extends AbstractAPIHandler implements A
 		JSONObject timeAndDate = new JSONObject();
 		
 		Date time_and_date = new Date();
-		timeAndDate.put("time_and_date", time_and_date.toString());
+		
+		String day = time_and_date.toString().substring(0, 3);
+		timeAndDate.put("day", day);
+		
+		String month = time_and_date.toString().substring(4, 7);
+		timeAndDate.put("month", month);
+		
+		String date = time_and_date.toString().substring(8, 10);
+		timeAndDate.put("date", date);
+		
+		String hours = time_and_date.toString().substring(11, 13);
+		timeAndDate.put("hours", hours);
+		
+		String minutes = time_and_date.toString().substring(14, 16);
+		timeAndDate.put("minutes", minutes);
+		
+		String seconds = time_and_date.toString().substring(17, 19);
+		timeAndDate.put("seconds", seconds);
+		
+		String timezone = time_and_date.toString().substring(20, 23);
+		timeAndDate.put("timezone", timezone);
+		
+		String year = time_and_date.toString().substring(24, 28);
+		timeAndDate.put("year", year);
 		
 		
 		JSONArray jsonArray = new JSONArray();


jig08:
Sample Link: [http://localhost:9000/api/console.json?q=SELECT%20*%20FROM%20timeanddate;](http://localhost:9000/api/console.json?q=SELECT%20*%20FROM%20timeanddate;)

screenshot from 2016-07-22 09 31 10

daminisatya:
@jig08 Please follow the review process i don't find an issue relevant to this PR. The issue which is being closed is no longer valid.

sudheesh001:
I believe this should have more functionality associated with it like for answering the questions which people might give to Susi
- What's the current time at `<location>`
- What's the time difference between `IST` and `GMT`
- What's the current date and time in `<location>`
- How many `<hours/minutes/seconds>` is it to `6:00PM` in `<location>`

People could phrase a lot of questions which are potentially dependant on time, We need to do the required computations than giving people a JSON which could potentially also be extracted from the [date object methods](http://www.w3schools.com/jsref/jsref_obj_date.asp).

jig08:
@daminisatya This wasn’t a major change an was very much related to #732 and #738.

Anyway, I have opened another issue for the same https://github.com/loklak/loklak_server/issues/747 :smiley:

jig08:
@sudheesh001 As mentioned https://github.com/loklak/loklak_server/issues/738#issuecomment-234362926 currently no third party service is being used.

> @daminisatya Currently, it aims to answer queries like:
    What is the time ?
    What is the date today ?
    Which day of the week is it ?
    What timezone are we in ? etc.

>For questions like:
   > Difference between timezones?
    What is the current time now?
    Timezone of a particular place?
    Date of a particular place?
    Time of a particular place?

> I would have to make use of a third party service like www.timeanddate.com or any other


If required, feel free to suggest any such service and open an issue for the same.

sudheesh001:
> no third party service is being used.

The aim is either to use the 3rd party service and scrape the required answers or to create the solution for answering the time queries, I would stick to the scraping of timeanddate.com and I think that's the right way to go about this.

Time is a very important thing in any assistant based system and needs to be handled correctly because a lot of potential operations could be based on these. Lets try and make this robust and handle different ways people can query.

jig08:
>I would stick to the scraping of timeanddate.com and I think that's the right way to go about this.

That was my plan too initially.

We can include system date/time for now.

And I will make a new console service for answering questions like : 
> 
    What's the current time at <location>
    What's the time difference between IST and GMT
    What's the current date and time in <location>
    How many <hours/minutes/seconds> is it to 6:00PM in <location>

jig08:
@sudheesh001 @daminisatya New issue ( #748 ) for parsing [timeanddate.com](http://www.timeanddate.com/) has been opened.

I beleive, for the time being we can use system date/time to make corresponding rules for _SUSI SU_

sudheesh001:
> That was my plan too initially.

So lets continue doing that and link this issue over there at #738 

> We can include system date/time for now.

Lets include all at once. Moving issue over to the new created issue #748

jig08:
Okay :+1:

shivenmian:
@sudheesh001 There's no need of using timeanddate.com. You can get date and time related information directly from java.util

sudheesh001:
What about the timezone differences ? what about time somewhere else ? Why would you want to compute all of that when you can go through timeanddate.com ?

Issue #740, opened 2016-07-21
Including a parser for Github profile


jig08:
Developing a console service by implementing a Github profile scraper

sudheesh001:
We can also get a lot of information from the publicly rate limited github APIs without the need of a key, Please look into that too.

jig08:
@sudheesh001 Yes. I was going through [developer.github.com/v3/rate_limit](https://developer.github.com/v3/rate_limit/) .

The rate_limited API can be easily integrated, should I go ahead and make a console service by using them or go the traditional way of parsing the HTML ?

jig08:
> Sudheesh Singanamalla @sudheesh001
The traditional way but have the fallback in case of failure to the regular API 

@sudheesh001 Sounds good. Shall make changes to PR #746  accordingly ...

jig08:
Resolved in PR #746

PR #746, opened 2016-07-22
Github Profile scraper

jig08:
... wrt https://github.com/loklak/loklak_server/issues/740

It scrapes accordingly for individual user account or an organization.

Sample Link: [http://localhost:9000/api/console.json?q=SELECT * FROM githubProfile WHERE profile='torvalds';](http://localhost:9000/api/console.json?q=SELECT%20*%20FROM%20githubProfile%20WHERE%20profile=%27torvalds%27;)
screenshot from 2016-07-22 11 24 47

Sample Link: [http://localhost:9000/api/console.json?q=SELECT * FROM githubProfile WHERE profile='loklak';](http://localhost:9000/api/console.json?q=SELECT%20*%20FROM%20githubProfile%20WHERE%20profile=%27loklak%27;)
screenshot from 2016-07-22 11 24 01

### DIFF
--- a/src/org/loklak/api/search/GithubProfileScraper.java
+++ b/src/org/loklak/api/search/GithubProfileScraper.java
@@ -1,0 +1,214 @@
+/**
+ *  Github Profile Crawler
+ *  Copyright 22.07.2016 by Jigyasa Grover, @jig08
+ *
+ *  This library is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public
+ *  License as published by the Free Software Foundation; either
+ *  version 2.1 of the License, or (at your option) any later version.
+ *  
+ *  This library is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *  Lesser General Public License for more details.
+ *  
+ *  You should have received a copy of the GNU Lesser General Public License
+ *  along with this program in the file lgpl21.txt
+ *  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.loklak.api.search;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import javax.servlet.http.HttpServletResponse;
+
+import org.json.JSONArray;
+import org.json.JSONObject;
+import org.json.JSONTokener;
+import org.jsoup.Jsoup;
+import org.jsoup.nodes.Document;
+import org.jsoup.nodes.Element;
+import org.jsoup.select.Elements;
+import org.loklak.server.APIException;
+import org.loklak.server.APIHandler;
+import org.loklak.server.AbstractAPIHandler;
+import org.loklak.server.Authorization;
+import org.loklak.server.BaseUserRole;
+import org.loklak.server.Query;
+import org.loklak.susi.SusiThought;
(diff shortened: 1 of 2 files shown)


src/org/loklak/api/search/GithubProfileScraper.java
```diff
@@ -0,0 +1,178 @@
...
+
+		String profileType = html.getElementsByAttributeValue("property", "og:type").attr("content");
+		githubProfile.put("type", profileType);
+
+		String username = html.getElementsByAttributeValue("property", "profile:username").attr("content");
+		githubProfile.put("username", username);
+
+		String description = html.getElementsByAttributeValue("property", "og:description").attr("content");
+		githubProfile.put("description", description);
+
```
sudheesh001: I am not sure if `og:` open graph would be the right way to do these, these open graph tags are generally kept so that facebook shares show up correctly and shouldn't be the representation of the values that are being scraped. The name and this kind of information is also available in the HTML without the `og:` tags

jig08: @sudheesh001 In my opinion, the information in `meta` with `og:` tags is the consistent information that can be scraped easily without going into the hassles of nested links.

Other scrapers like eventbrite, wordpress etc. have also been implemented on similar lines and have shown to be more steady in nature. 

What other way would you suggest else ?

jig08:
- Changed field names in result according to standard Github API
- In case of parsing/connection failure, it'll return results according to the rate limited Github API
- Sqaushed commits

src/org/loklak/api/search/GithubProfileScraper.java
sudheesh001: The OG tags and so on are used for SEO and not for showing content, scraping using these is not the right way to go about this and I think you need to scrape the actual HTML where the same content exists

jig08:
>The OG tags and so on are used for SEO and not for showing content, scraping using these is not the right way to go about this and I think you need to scrape the actual HTML where the same content exists

@sudheesh001 Changed accordingly.

src/org/loklak/api/search/GithubProfileScraper.java
```diff
@@ -0,0 +1,214 @@
...
+		Document html = null;
+
+		JSONObject githubProfile = new JSONObject();
+
+		try {
+			html = Jsoup.connect("https://github.com/" + profile).get();
+		} catch (IOException e) {
+			
+			URI uri = null;
```
sudheesh001: Is this step necessary can't we have URI defined in the same line as it is used ? since this `uri` variable isn't being used outside this catch.
`URI uri = new URI()` should be possible right ?

src/org/loklak/api/search/GithubProfileScraper.java
```diff
@@ -0,0 +1,214 @@
...
+			try {
+				uri = new URI("https://api.github.com/search/users?q=" + profile);
+			} catch (URISyntaxException e1) {
+				e1.printStackTrace();
+			}
+			
+			JSONTokener tokener = null;
+			try {
+				tokener = new JSONTokener(uri.toURL().openStream());
```
jig08: >Is this step necessary can't we have URI defined in the same line as it is used ? since this uri variable isn't being used outside this catch.
URI uri = new URI() should be possible right ?

@sudheesh001 Probably not. 
`uri` is being used here for `JSONTokener`

sudheesh001:
Works the way it is, needs implementation of #751

jig08:
Review++, hence merging this.

PR #749, opened 2016-07-22
#634: Scraper which scrapes a given weibo user's information

### LINKED ISSUE #634: [Weibo] Scrape user profile page and find profile user pages of followers and following

daminisatya:
Issue #634 : 
This PR aims at scraping the user's bio information from Weibo accounts. The following is the sample [URL](http://weibo.com/p/1005053472041784/info?mod=pedit_more) which is the bio information page. This patch needs some help with getting the source code from the above URL. 
If successful, the code can be tested properly.

### DIFF
--- a/src/org/loklak/api/search/WeiboUserInfo.java
+++ b/src/org/loklak/api/search/WeiboUserInfo.java
@@ -1,0 +1,138 @@
+/**
+ *  GenericScraper
+ *  Copyright 22.07.2016 by Damini Satya, @daminisatya
+ *
+ *  This library is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public
+ *  License as published by the Free Software Foundation; either
+ *  version 2.1 of the License, or (at your option) any later version.
+ *  
+ *  This library is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *  Lesser General Public License for more details.
+ *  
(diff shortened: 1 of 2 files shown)


daminisatya:
@DengYiping @sudheesh001 Please have a look at the problem. I'm unable to get the source code from the weibo urls. Jsoup alone is not able to get the html page. Is there any other possible way for doing this? Other than adding the personal browser cookies! 

@DengYiping Also please check on how we can queue this service, and use your patch into this. #735 

Thank you.

jig08:
@daminisatya The issue is that all the information resides inside the `<script>..</script>` tags. 
The way out is to parse the javascript code. 

As mentioned [here](http://stackoverflow.com/a/14925848), "JSoup + Manual Parsing" seems like a good and a viable option.

DengYiping:
#735 This is the helper function implementing the idea of @jig08

DengYiping:
@daminisatya How's it going on this PR? Anything I can help?

src/org/loklak/api/search/WeiboUserInfo.java
```diff
@@ -0,0 +1,138 @@
...
+	@Override
+	protected void doGet(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
+		Query post = RemoteAccess.evaluate(request);
+
+		// manage DoS
+		if (post.isDoS_blackout()) {response.sendError(503, "your request frequency is too high"); return;}
+		
+		String url = post.get("url", "");
+		JSONObject obj = new JSONObject();
+		Document doc = Jsoup.connect(url).get();
```
DengYiping: @daminisatya Here, you should use the helper function.

Issue #748, opened 2016-07-22
Parse third party service (timeanddate.com) for expanding scope of SUSI


jig08:
To make SUSI answer questions like :

- [ ]  What's the current time/day/date ? 
- [x]  What's the current time at `<location>` ?
- [ ]  What's the time difference between `IST` and `GMT` ?
- [ ]  How many `<hours/minutes/seconds>` is it to `6:00PM` in `<location>` ?

Parse third party services like [timeanddate.com](http://www.timeanddate.com/).

Follow up of Issues #732 , #747 and PRs #738 #745

jig08:
Suggested alternative instead of [timeanddate.com](http://www.timeanddate.com/) :- 
 [stackoverflow.com/questions/589870/should-i-use-java-date-and-time-classes-or-go-with-a-3rd-party-library-like-joda](http://stackoverflow.com/questions/589870/should-i-use-java-date-and-time-classes-or-go-with-a-3rd-party-library-like-joda)

PR #715, opened 2016-07-18
Adding time and date console service.

jig08:
Test link: [http://localhost:9000/api/console.json?q=SELECT * FROM timeanddate WHERE url='http://www.timeanddate.com/';](http://localhost:9000/api/console.json?q=SELECT%20*%20FROM%20timeanddate%20WHERE%20url=%27http://www.timeanddate.com/%27;)

screenshot from 2016-07-18 20 33 48

PR #774, opened 2016-07-31
Service for current date and time

jig08:
... in partial fulfillment of issue #748 

Sample Link: [http://localhost:9000/api/console.json?q=SELECT * FROM currentdateandtime;](http://localhost:9000/api/console.json?q=SELECT%20*%20FROM%20currentdateandtime;)

screen shot 2016-07-31 at 4 43 30 pm


DengYiping:
@jig08 It is a good idea to implement time zone offset also.

jig08:
@DengYiping You meant offset from UTC ? 
If yes, I am not sure it can be done without using any third-party service. In this case, the system date-time have been extracted.

sudheesh001:
The question here would be why would one need this as a JSON when it's easier to obtain from a javascript core functionalities normally, Making a request and then fetching time seems crazy to me. Can you provide a use case

jig08:
@sudheesh001 Was just making a service on lines of others like `location-wise time` for SUSI to resolve issue #748

>To make SUSI answer questions like :
 What's the current time/day/date ?
 What's the current time at <-location-> ?
 What's the time difference between IST and GMT ?
 How many <-hours/minutes/seconds-> is it to 6:00PM in <-location-> ?

Agreed it can be done via Javascript core functionalities though, go ahead and include it in Susi's capabilities :+1:

sudheesh001:
> What's the current time/day/date ?
What's the current time at <-location-> ?
What's the time difference between IST and GMT ?
How many <-hours/minutes/seconds-> is it to 6:00PM in <-location-> ?

We can get this information even from timeanddate.com right ? Why is there the problem ?

jig08:
@sudheesh001 The last time I had done the current date-time using timeanddate.com I had been asked not to use third party service for this. 
Kindly follow up here https://github.com/loklak/loklak_server/pull/715

Issue #519, opened 2016-06-03
Installation page


treba123:
We're getting more and more features that can or should be customized by the admin of a loklak installation.
Maybe we should get a simple installation page sooner or later, going through the settings on first run, like many php projects like wordpress do.
Shouldn't be to much work and can make it much easier to discover certain features

I'm thinking things like:
- send mails via smtp (email address, server, login)
- installation name
- admin name and password (as soon as we have real accounts)
- backend server
- https settings (are you behind a ssl-proxy?)
- ...

shivenmian:
Yes.

Orbiter:
+1
I would consider that this can be seen in the context oft personal settings for each user where the admin has more setting options, most oft them you already mentioned.

treba123:
@shivenmian @Orbiter @mariobehling The installation page is now functional, but i've got some questions before we can merge it:

1. Should the installation page be displayed by default (on first run)? 
  - One problem with that would be the one-click deployments, as one would have to make sure to complete the installation so no stranger does it.
  - What happens with installations that just do an upgrade? Should they also be forced to make the installation? Would probably make sense.

2. Should the installation page provide a 'restart' button, that reloads the config and restarts all services? Or should the user have to do that himself? I'd say a restart button would be helpful for beginners.

3. Should the installation service be the only service being started at first run? That would make sure everything is set up before normal operation begins.

4. Would you like to have additional features?

The current features are:
- admin account creation
- a user registration setting
- general settings like host- and shortlink url, peername, backend settings
- smtp settings (with test option)
- https settings

mariobehling:
How does the page look like? Any screenshot?

mariobehling:
I am not clear on your idea of the implementation, but as of now the service always restarted itself after an update/upgrade. And as loklak is constantly providing updates/upgrades the goal was always to provide upgrades that work without the need of any user actions. Do I understand you correctly, that this is no longer possible in the way you propose it here?

treba123:
It's totally possible. The question is, how do we want it. If you say installation should not happen on upgraded installations, I'll try to make it that way.

Here are some fast screenshots:

![installation0](https://cloud.githubusercontent.com/assets/6313774/17365089/75d14238-5984-11e6-9679-f498b17071f1.png)
![installation1](https://cloud.githubusercontent.com/assets/6313774/17365088/75cd2d60-5984-11e6-9282-0ae15f36d79b.png)
[...]

PR #775, opened 2016-08-01
Installation page

treba123:
**@ Codacy fail:**
Codacy is failing on this PR as libs like jquery get reviewed, too, which do not pass the test. The code contributions from me are fine.

This is a working version of the installation page https://github.com/loklak/loklak_server/issues/519
Warning: requires bash for shell interaction!

It currently offers:
- admin account creation
- a user registration setting
- general settings like host- and shortlink url, peername, backend settings
- smtp settings (with test option)
- https settings

The installation is started with an own start script, in it's own main class. Launching it will be offered by the normal start script.
In the process I also made some changes/additions to the email handler

How to test:
- clonegit@github.com:treba123/loklak_server.git
- checkout development
- build and start via start.sh
- choose to do the installation
- go through the dialog
- afterwards, the normal server should be started

**Codacy:**
I adopted most recommendations, most remaining issues are with external files which I have little influence on.


mariobehling:
> 3. what happens on already deployed servers that just get updated? (currently the installation gets disabled in the customized_config.properties, but we can't set this value on upgrade)
4. should the installation provide a restart button (needed for some settings to take effect) or should the user do that manually?

As loklak is constantly providing updates/upgrades the goal was always to provide upgrades that work without the need of any user actions. Do I understand you correctly, that this is no longer possible in the way you propose it here?

mariobehling:
Please disregard my previous comment and let's move this discussion to the issue itself #519

daminisatya:
@treba123 is this out for review?

treba123:
@daminisatya yep :) was just about to ask on gitter.

src/org/loklak/LoklakInstallation.java
```diff
@@ -0,0 +1,478 @@
+/**
+ *  LoklakServer
+ *  Copyright 22.02.2015 by Michael Peter Christen, @0rb1t3r
```
daminisatya: @treba123  I guess you need to change this line.

shivenmian:
@treba123 Please squash before you merge :)

treba123:
Will do :)

treba123:
Hm somehow the commit fails to build, but runs fine on my mashine. Any ideas @all ?

treba123:
@sudheesh001 this pr also includes some changes to the login/signup pages as I had to make some changes on the backend and therefor went ahead to also make some small changes on the frontend aswell. 
They don't change anything fundamentally but make use of native browser functions, which mostly benefits touch devices and reduce code complexity a bit.
Hope it's alright.

sudheesh001:
screen shot 2016-08-11 at 10 41 39 am

Button breaks and are not clickable, keeping the home page at the installation wizard

sudheesh001:
Hitting (r) Remind me next time also takes me to the installation wizard page,it should instead take me to the home page.

treba123:
@sudheesh001 the password is not not sufficient. I'll make the hint for that more prominent.
About hitting r, what's the output of the start script? Does it report 'starting loklak' or 'starting loklak installation' ?
Because I noticed that quite often if you first started the installation, the browser does not refresh afterwards. Could you check this by emtying the browser cache or using a second browser?

Thanks for testing!

treba123:
@sudheesh001 I think I found a solution for the browser caching issue, so the issues you faced should be solved. Could you retest at some point?

treba123:
Relabel to 'requesting review' as the issues encountered should be fixed.

jig08:
@treba123 `codacy` checks fail, kindly review

treba123:
@jig08 this is not because of my code but of other libs we're using (like jquery). See first post.
They should not get reviewed.

sudheesh001:
Reviewing.

sudheesh001:
screen shot 2016-08-17 at 4 39 46 pm
Same issue with the installation page.
What is going well
[x] Hitting `n` takes to loklak home page
[x] Remind me next time seems to be working

What is not going well
[ ] Hitting the continue button after filling the form items
[ ] There should be a UI Button to also `Skip Installation` from the Web UI.

treba123:
@sudheesh001 well, there is already the abort button, which ends the installation. I don't see why there would be a need for a skip button there, as to start the installation you have to explicitly type 'y' in the script. In what case would that be of benefit?

At the password thing: are you really sure your password meets the requirements as written in the hint? Because I tested it on chrome here and it works fine...

sudheesh001:
Abort is generally used in places where you'd like to Stop the complete execution, Skip refers to skip a step. Generally speaking the Abort should stop the loklak server in the installation, similar to how hitting Abort/Exit should stop any installation process that's going on. So the Abort and Skip are two different functionalities altogether. I hope it's more clear now

treba123:
Ah alright. But then 'skip installation' does not make sense here (IMO), as the installation is a whole process in itself. So the 'abort' refers to the installation, which can also be launched on it own and is not necessarily part of the normal start procedure.

sudheesh001:
The functionality looks okay but I really think we need a `Skip Install for now / Remind me later` button on the UI itself. I think @Orbiter should make the review.

Orbiter:
Hi, that is good work, however there are some things which did not work when I tested this:

- startup of bin/installation.sh and entering of the installation detailes worked smoothly. Very good work! One button on the first page looks a bit shifted to the side, but whatever
- after the shutdown, I started normally but loklak complained that the installation was not done:
```
$:loklak_server admin$ bin/installation.sh
starting loklak installation
loklak installation started at port 9000, open your browser at http://localhost:9000
$:loklak_server admin$ bin/start.sh
Loklak detected that you did not yet run the installation wizard.
It let's you setup an administrator account and a number of settings, but is not mandatory.
You can manually start it by running bin/installation.sh
Would you like to start the installation now? (y)es, (n)o, (r)emind me next time
```
However, while this message was shown, the server started up nevertheless.

I also checked the customized_config.properties and found some problems:
```
backend=http\://loklak.org
host.url=http\://127.0.0.1\:9000
shortlink.urlstub=http\://127.0.0.1\:9000
```
These urls look broken.
Only the last thing looks like a bug, the other things are more or less cosmetics. Can you please have a look for the strange urls?

treba123:
@Orbiter thanks for the review!
The urls are actually right, I wondered about this, too. Here is an [explanation](https://docs.oracle.com/javase/7/docs/api/java/util/Properties.html#store(java.io.Writer,%20java.lang.String)) 'The key and element characters #, !, =, and : are written with a preceding backslash to ensure that they are properly loaded.'
When loaded by the Loklak main class, the urls a fine again.

About the installation reminder: right, I missed that. Will fix :)

Orbiter:
Strange thing with the properties encoding but ok.

About the recently discussed abort option: I don't think we need that on the web page. But we would require some transparency about the whole process when the installation wizard starts, like:
- print on the terminal that the server terminates after the wizard was run successfully
- have the option to stop the wizard with bin/stop.sh (I would guess that will work anyway?)

Issue #771, opened 2016-07-30
Fresh clone from Master wont start.


smokingwheels:
Fresh clone from Master 3 times tested.
Log file Zero bytes.
No rush yet.
Minor spelling error no problem.

BUILD SUCCESSFUL
Total time: 7 seconds
root@Debian:/usr/loklak_server# bin/start.sh
starting loklak
cat: data/startup.tmp: No such file or directory
bin/start.sh: 47: [: =: unexpected operator
loklak server failed to start. See data/loklag.log for details
root@Debian:/usr/loklak_server#

jig08:
@smokingwheels I tried cloning, building and running the `master` branch of Loklak and it works just fine for me.

```
jigyasa@jigyasa-HP-Pavilion-15-Notebook-PC:~/masterLoklak$ git clone https://github.com/loklak/loklak_server
Cloning into 'loklak_server'...
remote: Counting objects: 43272, done.
remote: Compressing objects: 100% (33/33), done.
remote: Total 43272 (delta 7), reused 0 (delta 0), pack-reused 43238
Receiving objects: 100% (43272/43272), 197.85 MiB | 2.78 MiB/s, done.
Resolving deltas: 100% (34929/34929), done.
Checking connectivity... done.
jigyasa@jigyasa-HP-Pavilion-15-Notebook-PC:~/masterLoklak$ cd loklak_server/
jigyasa@jigyasa-HP-Pavilion-15-Notebook-PC:~/masterLoklak/loklak_server$ git branch -a
* development
  remotes/origin/HEAD -> origin/development
  remotes/origin/development
  remotes/origin/gh-pages
  remotes/origin/master
jigyasa@jigyasa-HP-Pavilion-15-Notebook-PC:~/masterLoklak/loklak_server$ git checkout master 
Branch master set up to track remote branch master from origin.
Switched to a new branch 'master'
jigyasa@jigyasa-HP-Pavilion-15-Notebook-PC:~/masterLoklak/loklak_server$ ant
Buildfile: /home/jigyasa/masterLoklak/loklak_server/build.xml

init:

build:
    [mkdir] Created dir: /home/jigyasa/masterLoklak/loklak_server/classes
     [echo] loklak: /home/jigyasa/masterLoklak/loklak_server/build.xml
    [javac] Compiling 183 source files to /home/jigyasa/masterLoklak/loklak_server/classes
    [javac] Note: Some input files use or override a deprecated API.
[...]

PR #776, opened 2016-08-02
Log all output to data/loklak.log

treba123:
There should only be output from the console when the log-infrastructure fails. We can therefor write it into the same log file.
This should it make easier for users to spot issues.

(no issue for this, but there were multiple issues already where people didn't know about the second log file, for example https://github.com/loklak/loklak_server/issues/771)


treba123:
@daminisatya thank for the fast review :) did you start the server once with this change, just to check if anything unexpected happens?

sudheesh001:
@treba123 Looks like no problems in the functionality.

treba123:
@sudheesh001 thanks!

PR #780, opened 2016-08-03
Fixes #779: changing visibility to public of GenericJSONBuilder

### LINKED ISSUE #779: Error while using DAOWrapper
Error faced while using the `DAOWrapper ` class implemented in PR #772 
`The type DAOWrapper.GenericJSONBuilder is not visible`

Fix is to make it public/visible ?

.. Related to issue #751

jig08:
... wrt #779


jig08:
Not a major change, merging hence.

Issue #783, opened 2016-08-03
Add OpenJDK to travis


treba123:
OpenJDK is the java reference implementation, therefor to avoid usage of non-standard oraclejdk specific features (as happened in https://github.com/loklak/loklak_server/issues/782), we should include OpenJDK in the travis build process.

treba123:
Had to revert #784 (https://github.com/loklak/loklak_server/pull/811) because of run errors.
We can try again in some time when the support is more stable.

Issue #782, opened 2016-08-03
JavaFX not bundled with OpenJDK


treba123:
https://github.com/loklak/loklak_server/commit/2f8184b2ba1c1c7dc4c48f051aac3182bec0a5ef imports JavaFX, which is fine for oracleJDK, which bundles it with jdk8, but not for openjdk, which is used by many linux distributions.

As the used functionality is rather trivial, we should better replace that lines.

DengYiping:
@treba123 Thanks for mention. Fix is here: #788

PR #784, opened 2016-08-03
 Add openjdk8 to travis (to ensure portability) 

treba123:
see https://github.com/loklak/loklak_server/issues/783 https://github.com/loklak/loklak_server/issues/782

Edit2: reverted the start script changes. they will come together with the installation page.

Edit: made some adjustments on the start script. It should now fail correctly if loklak can't be launched at all.
Before merging please test what happens when you do
- ant clean
- bin/start.sh


DengYiping:
@treba123 Please squash commit. Wait for #788 to be merged, so that you can pass travis-ci test.Cheers!

.travis.yml
```diff
@@ -1,6 +1,9 @@
 language: java
+sudo: required
```
sudheesh001: Is there a reason why we need sudo or is it just to make the builds faster ?

treba123:
Without sudo, we can't use trusty for some reason. Tried it without, didn't work.

.travis.yml
treba123: Yes, trusty only works with it. And we need trusty for openjdk8

PR #788, opened 2016-08-04
Fixes #782 adds TwoTuple class and removes dependency on Javafx(OracleJDK)

### LINKED ISSUE #782: JavaFX not bundled with OpenJDK

DengYiping:


DengYiping:
#Fix #782

sudheesh001:
@DengYiping This looks good, can you please amend the commit message as 
` Fixes #782 adds TwoTuple class and removes dependency on Javafx(OracleJDK)`

This is then good to merge

sudheesh001:
taking over, editing and merging so that https://github.com/loklak/loklak_server/pull/784 is unblocked

PR #800, opened 2016-08-06
Fixes #750: Implement Amazon API

### LINKED ISSUE #750: Propose and implement Amazon API
There are product APIs (https://affiliate-program.amazon.com/gp/advertising/api/detail/main.html) and a number of other APIs (https://developer.amazon.com/). Which ones could be relevant for Susi?

* [x] List relevant APIs in this issue
* [ ] As a start implement an API that offers information on products

shivenmian:
@mariobehling @Orbiter 

I have implemented one part of the Amazon Product Advertising API integration, the ItemLookup API. 

Supply an ASIN / Product ID, and if the product is in the database (and if it's approved by it's parent company to be displayed by the API), it gives you necessary info about the result.

For searches like Apple Products (I did a search on iPhone), it isn't available in the Advertising API, because Apple does not want to display the prices of its products that way or something.

If you wish to test, please contact me. I have the AWS Keys and the Associate Tags, which I can't display here right now. I think these fields should be supplied as a GET parameter instead.

TODO:

* [x] Make it more efficient. The Amazon API is a SOAP API, not REST. Some parameters are available in some products, some aren't.

* [x] Have a feature to list other variants of the product, and possibly list the cheapest variant too.

* [x] Add the SimilarityLookup and ItemSearch API (in ItemSearch, you have Search by Product Name).

Screenshot: 

screen shot 2016-08-06 at 5 01 01 pm

### DIFF
--- a/src/org/loklak/api/amazon/AmazonProductService.java
+++ b/src/org/loklak/api/amazon/AmazonProductService.java
@@ -1,0 +1,162 @@
+/**
+ *  AmazonProductService
+ *  Copyright 05.08.2016 by Shiven Mian, @shivenmian
+ *
+ *  This library is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public
+ *  License as published by the Free Software Foundation; either
+ *  version 2.1 of the License, or (at your option) any later version.
+ *  
+ *  This library is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *  Lesser General Public License for more details.
+ *  
+ *  You should have received a copy of the GNU Lesser General Public License
+ *  along with this program in the file lgpl21.txt
+ *  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.loklak.api.amazon;
+
+import java.io.StringWriter;
+
+import javax.servlet.http.HttpServletResponse;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
+
+import org.json.JSONObject;
+import org.json.XML;
+import org.loklak.data.DAO;
+import org.loklak.server.APIException;
+import org.loklak.server.APIHandler;
+import org.loklak.server.AbstractAPIHandler;
+import org.loklak.server.Authorization;
+import org.loklak.server.BaseUserRole;
+import org.loklak.server.Query;
+import org.loklak.tools.storage.JSONObjectWithDefault;
+import org.w3c.dom.Document;
+
+public class AmazonProductService extends AbstractAPIHandler implements APIHandler {
+
+	private static final long serialVersionUID = 2279773523424505716L;
+
+	// set your key configuration in config.properties under the Amazon API
+	// Settings field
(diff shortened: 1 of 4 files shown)


sudheesh001:
Most Amazon products have a table describing the item dimensions etc.., This can be converted to the JSON object too, Can you look into that, It doesn't look like there's a lot of information being collected here.

shivenmian:
@sudheesh001 Yes, the ItemDimensions field was definitely there, I just separated it into Height, Width and Length as you saw.

And yes, there's more to be done, as I mentioned in the TODO.

sudheesh001:
@shivenmian I meant tables like this one
https://www.amazon.com/gp/product/B00OQVZDJM/ref=s9_acss_bw_cg_odsbnc_1b1?pf_rd_m=ATVPDKIKX0DER&pf_rd_s=merchandised-search-2&pf_rd_r=BDM863CM55C13G3866V8&pf_rd_t=101&pf_rd_p=7f7be4a4-f096-4b82-aa90-06a6944c0436&pf_rd_i=6669702011

screen shot 2016-08-06 at 5 38 19 pm

shivenmian:
Oh. Didn't think of this one. Will look into it.

shivenmian:
@mariobehling @sudheesh001 Have completed the Integration.

**API usage:**

Params you need to supply:

1. Amazon Item ID (**id**) or Item Name (**q**). Either of them but one should be there.
2. [Response Group](http://docs.aws.amazon.com/AWSECommerceService/latest/DG/CHAP_ResponseGroupsList.html)(**response_group**). This is optional. If this is not supplied, Response Group is defaulted as Large, and Large gives all the possible info.

src/org/loklak/api/amazon/AmazonProductService.java
```diff
@@ -0,0 +1,160 @@
...
+
+public class AmazonProductService extends AbstractAPIHandler implements APIHandler {
+
+	private static final long serialVersionUID = 2279773523424505716L;
+
+	// keys available on request
+	private static final String AWS_ACCESS_KEY_ID = "";
+	private static final String AWS_SECRET_KEY = "";
+	private static final String ASSOCIATE_TAG = "";
+
```
sudheesh001:
```java
	private static final String AWS_ACCESS_KEY_ID = "";
	private static final String AWS_SECRET_KEY = "";
	private static final String ASSOCIATE_TAG = "";
```

These are configurable entries, can you please move them to the `conf` file and use the string fields from there ? We wouldn't expect a user to open a java file and make changes to the KeyID and the Key / Tags

shivenmian: Makes sense, hmm.

src/org/loklak/api/amazon/AmazonProductService.java
```diff
@@ -0,0 +1,160 @@
...
+			TransformerFactory tf = TransformerFactory.newInstance();
+			Transformer transformer = tf.newTransformer();
+			transformer.transform(domSource, result);
+			JSONObject xmlresult = new JSONObject(true);
+			xmlresult = XML.toJSONObject(writer.toString());
+			JSONObject items = xmlresult.getJSONObject(operation).getJSONObject("Items");
+			if (items.getJSONObject("Request").has("Errors")) {
+				itemlookup.put("status", "error");
+				itemlookup.put("reason",
+						items.getJSONObject("Request").getJSONObject("Errors").getJSONObject("Error").get("Message"));
```
sudheesh001: Will the fetch always be in this format ?

shivenmian: Yes. I am implementing the ItemLookup and the ItemSearch API, and they are always in this format.

There are other APIs too: like SimilarityLookup etc, but the result of SimilarityLookup (and all the other APIs) comes when ```response_group=Large```, and as I wrote it's that by default in my code, so if you want all the results, just fetch an ```id``` or a ```q``` and leave response_group blank.

PR #819, opened 2016-08-13
Fixes #807: Handling the internal server error

### LINKED ISSUE #807: [Scrapers] The invalid URLs must be handled gracefully
As of now the scrapers which are implemented are able to scrape data for a particular URL formats. If in case there is an unsupported input, the output should be handled gracefully instead of throwing a ```server error``` 

Here's a sample erroneous output on using genericscraper. 
selection_276

This is the case with every scraper, there must be a meaningful error message along with the expected url format.

daminisatya:
#807 
This patch will gracefully handles the server errors, throwing a meaningful error message.

### DIFF
--- a/src/org/loklak/LoklakErrorHandler.java
+++ b/src/org/loklak/LoklakErrorHandler.java
@@ -24,6 +24,7 @@ import java.io.Writer;
 
 import javax.servlet.http.HttpServletRequest;
 
+import org.json.JSONObject;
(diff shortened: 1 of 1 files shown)


daminisatya:
selection_281

daminisatya:
@sudheesh001 Please help me with the error message. Is it ok to display the JSONObject? or a plain text message?

shivenmian:
+1

Issue #448, opened 2016-05-26
Make signatures on JSONObjects


treba123:
For #378 it would be nice to have signatures on the every timeline as also on every other json object if needed

Orbiter:
Please implement this.

treba123:
Alright, will work on it tomorrow

treba123:
Alright, so from what I understand, there is a standard for signing json objects called [jose](http://jose.readthedocs.io/en/latest/), but it is much much bigger than what we actually need.

So for now I will implement a much simpler method which can directly used without interfering with any other loklak component.
It will as simple as adding a `signature` field to the object, which contains a signature of the sha256 hash of the object (without the signature itself of course).

The main issue will probably be to make sure the hash is deterministic (that is, as json objects are not necessarily ordered, the hash must always be the same as long as the content is the same)

treba123:
Implementation here: https://github.com/treba123/loklak_server/commits/signatures
Needs some more testing if the signature stays valid under various circumstances.

@Orbiter should I add this to the timelines? The creation of signatures and the verification?

Orbiter:
Is there a lot of computation required, or space? If not, give it a try.

Orbiter:
Is there a lot of computation required, or space? If not, give it a try.

treba123:
https://github.com/loklak/loklak_server/pull/837

PR #837, opened 2016-08-22
Signatures on JSONObjects

treba123:
This implements https://github.com/loklak/loklak_server/issues/448 and adds signatures to timelines. A propper verification of the timeline requires more work on peers (remembering peers and their keys).


sudheesh001:
@treba123 This looks good, can you squash please 😄

shivenmian:
@Orbiter @sudheesh001 I think the commits should be squashed.

treba123:
@sudheesh001 @shivenmian you're right, should have been squashed, sorry. But thanks for having a look!

PR #850, opened 2016-10-28
Declare project name for Gradle

yukiisbored:
It's a good idea to declare a project name for Gradle so it doesn't
assume the working directory (the directory where the local git repo
is at) as the project name.

PR #851, opened 2016-10-28
Define source and target compability to 1.8

yukiisbored:
In the Ant build.xml, It defines the source and target compatibility to
1.8. For consistency sake, It's a good idea to set the same thing on
Gradle as well

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,8 @@
 apply plugin: 'java'
 
+sourceCompatibility = 1.8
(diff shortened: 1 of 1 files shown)


shivenmian:
It'd be great if you send one squashed PR with all the changes for #850, #851 and #852, not separate PRs with changes (since you sent these PRs at the same time and edited the same file).

Make PRs great again :)

PR #856, opened 2016-10-29
Fixes #854 Updates ant build to include all jars in lib/ into classpath

### LINKED ISSUE #854: Ant: Include complete lib/ directory for  jars instead of manually including jars.
Currently the build consists of a build file with all the jars present in them manually listed.
This causes it to be bulky and creates a lot of errors among new developers because they either forget to add a jar or something similar.

The best thing to do would be to remove the manual listing and include all the jars `**.*.jar` from the directory into the build file.

sudheesh001:


shivenmian:
Good one @sudheesh001

PR #852, opened 2016-10-28
Add Gradle generated classes to the classpath

yukiisbored:
Some people may prefer to use Gradle over Ant to build Loklak so It's a
good idea to add Gradle's generated classes to the classpath of start.sh
and start.bat

PR #877, opened 2016-12-09
Add Gradle generated classes to the classpath

yukiisbored:
Some people may prefer to use Gradle over Ant to build Loklak so It's a
good idea to add Gradle's generated classes to the classpath of start.sh
and start.bat

Fixes https://github.com/loklak/loklak_server/issues/852


yukiisbored:
@mariobehling Here you go.

Issue #873, opened 2016-11-30
Add Gradle generated classes to classpath for start/stop scripts


yukiisbored:
The start/stop scripts on Loklak doesn't use Gradle generated classes at all. It only uses the ones generated by Ant. Gradle generated classes should be added because not everyone have Ant or want to use it.

jayvdb:
Note this is solved by https://github.com/loklak/loklak_server/pull/852 . I presume that makes it ineligible for GCI due to anti-plagiarism rules at least. (I'm still seeking clarification regarding prior work for a different org at https://www.mediawiki.org/wiki/Talk:Google_Code-in_2016/Mentors )

PR #885, opened 2016-12-11
Add Gradle generated classes to classpath

yukiisbored:
Some of the scripts doesn't include the Gradle generated classes to
the classpath. This commit adds the classes and checks the
availability of ant or gradle generated classes.

Closes https://github.com/loklak/loklak_server/issues/873


niccokunzmann:
@yukiisbored Do you know: if this was bash, you could put all these into one file and use the source command. I am not sure if that is possible with sh. Having it in one file is a great relief for maintenance.

yukiisbored:
@niccokunzmann Yeah, I already know about that. I'll probably do that later since it's kinda risky

Issue #886, opened 2016-12-12
Decrease redundant/duplicate code on scripts


yukiisbored:
Loklak has a number of scripts that have redundant/duplicate code on the scripts. This decreases maintainability. The solution to do this is to separate the same instructions to it's own script and run them via the `source` instruction.

PR #887, opened 2016-12-12
Reduce redundancy and code duplication on shell scripts

yukiisbored:
The shell scripts which are used to start/stop Loklak has a huge
amount of duplicate code. This commit reduces the amount of redundancy
and duplication by separating them into it's own script.

Fixes https://github.com/loklak/loklak_server/issues/886

### DIFF
--- a/bin/.preload.sh
+++ b/bin/.preload.sh
@@ -30,3 +30,27 @@ if [ -f $CUSTOMCONFIG ]; then
     j="$(grep Xmx $CUSTOMCONFIG | sed 's/^[^=]*=//')";
     if [ -n $j ]; then CUSTOMXmx="$j"; fi;
 fi
+
+CLASSPATH=""
+for N in lib/*.jar; do CLASSPATH="$CLASSPATH$N:"; done
+
+if [ -d "./classes" ]; then
+    CLASSPATH=".:./classes/:$CLASSPATH"
+elif [ -d "./build/classes/main" ]; then
+    CLASSPATH=".:./build/classes/main:$CLASSPATH"
+else
+    echo "It seems you haven't compile Loklak"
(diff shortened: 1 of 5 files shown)


daminisatya:
@sudheesh001 Can you please have a look at this ?

niccokunzmann:
This is a lot of code which you change. If you like to make a reviewer happy, you can try this in one PR each, notifying me and I can give you input

- Create the file where the duplicate code should go to, source this file in every bash script
- For each small duplication do:
    - delete from source file, put it into the new file

Also, you may want to look into functions, so the architecture can be decoupled from the use-cases (which does not need to be a good thing). Functions may make the code-reuse more obvious.

yukiisbored:
Ok

yukiisbored:
@niccokunzmann I've seperated the changes into multiple commits. I hope it's good to review.

niccokunzmann:
I do not understand the last commit but I would merge the other commits. They are a joy to read. Really obvious what happens there.

yukiisbored:
@niccokunzmann The last commit basically remove redundant code from `heroku_start.sh` and `bluemix_start.sh` by using `start.sh` with a couple of newly defined variables. Want me to revise the commit message?

niccokunzmann:
If you remove the last commit, I will merge the ones before. Then, we can look at the last commit.
If you like: you call start.sh - It is clearer to pass arguments to start.sh than to crate variables for it.
This way, one can be sure that the arguments are only used in start.sh.

yukiisbored:
@niccokunzmann I've changed from using variables to command arguments. I'm forced to use short single-characters flags because `getopt` is not available on macOS :(

niccokunzmann:
Very nice! With getopt. I think, I merge it it looks good and a lot of work went into it.

PR #907, opened 2016-12-26
Minify Docker image

yukiisbored:
Scripts are modified to be able to run on other UNIX utilities than GNU
coreutils, since Alpine uses BusyBox.

Docker Image has been minified by replacing the distribution and
removing all of the unneeded packages.

Docker Image building and testing is now part of Travis

### Short description
I have:

### For the reviewers
I have:

### DIFF
--- a/bin/.preload.sh
+++ b/bin/.preload.sh
@@ -14,7 +14,7 @@ mkdir -p data/settings
 #to not allow process to overwrite the already running one.
 if [ -f $PIDFILE ]; then
     PID=$(cat $PIDFILE 2>/dev/null)
-    if [ $(ps -p $PID -o pid=) ]; then
+    if kill $PID > /dev/null 2>&1; then
         echo "Server is already running, please stop it and then start"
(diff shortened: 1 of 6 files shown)


yukiisbored:
For those who don't know, Kill signal 0 means `CANCEL`, It doesn't do anything and can be used to check the alive status of a process. GNU's `ps` options is not part of the `POSIX` standard so it's much better to use `kill -0` than a fancy `ps` check.

yukiisbored:
Travis will fail, because the changes of the scripts are inside this PR.

bin/.preload.sh
```diff
@@ -12,9 +12,9 @@ CUSTOMXmx=""
 mkdir -p data/settings
 
 #to not allow process to overwrite the already running one.
-if [ -f $PIDFILE ]; then
+if [ -f $PIDFILE ] ; then
```
jayvdb: inconsistent style.

Issue #913, opened 2016-12-28
org.loklak.server.Query isn't working properly for HTTP POST


yukiisbored:
It looks like `org.loklak.server.Query` isn't working properly for HTTP POST

**HTTP `GET`**:
`yuki@matsuoka ~ $ curl -X GET -i 'http://loklak.org/api/search.json?q=loklak&source=twitter'`
```
HTTP/1.1 200 OK
Server: nginx/1.6.2
Date: Wed, 28 Dec 2016 08:02:37 GMT
Content-Type: application/json;charset=utf-8
Content-Length: 24989
Connection: keep-alive
Set-Cookie: JSESSIONID=p165k4rlo9ax1tz5g3fz55zpt;Path=/
Expires: Wed, 28 Dec 2016 08:02:36 GMT
Last-Modified: Wed, 28 Dec 2016 08:02:36 GMT
X-Robots-Tag: noindex,noarchive,nofollow,nosnippet

{
  "readme_0": "THIS JSON IS THE RESULT OF YOUR SEARCH QUERY - THERE IS NO WEB PAGE WHICH SHOWS THE RESULT!",
  "readme_1": "loklak.org is the framework for a message search system, not the portal, read: http://loklak.org/about.html#notasearchportal",
  "readme_2": "This is supposed to be the back-end of a search portal. For the api, see http://loklak.org/api.html",
  "readme_3": "Parameters q=(query), source=(cache|backend|twitter|all), callback=p for jsonp, maximumRecords=(message count), minified=(true|false)",
  "search_metadata": {
    "startRecord": "1",
    "maximumRecords": "20",
    "count": "14",
    "hits": 14,
    "period": 25434001,
    "query": "loklak\\",
    "client": "139.195.62.119",
    "time": 933,
    "servicereduction": "false",
    "count_twitter_all": 0,
    "count_twitter_new": 14,
    "count_backend": 0,
    "cache_hits": 0,
    "scraperInfo": "local"
  },
  "statuses": [
    {
      "timestamp": "2016-12-28T08:02:36.917Z",
      "created_at": "2016-12-27T10:24:41.000Z",
      "text": "@0rb1t3r We have wind. Maybe get a kite? We have an emergency here now. Are any @yacy_search/@loklak_app events planned in Germany ....
```

**HTTP `POST`**

`yuki@matsuoka ~ $ curl -X POST -i 'http://loklak.org/api/search.json' -d 'q=loklak&source=twitter'`
```
[...]

Issue #919, opened 2016-12-29
Implement an admin role to change the configuration on the settings page


mariobehling:
The settings page already enables changes on localhost. Please also implement the ability of an admin on systems on the Internet.

* [ ] Use the account apps to administer roles
* [ ] Make settings viewable to anonymous users only
* [ ] Make settings editable by admin users
* [ ] Merge the dumps page in the settings page

mariobehling:
The account apps were taken out some time ago when we moved the "apps" to http://apps.loklak.org. Please re-add the account apps in order to proceed. Compare to http://apps.loklak.org/#AccountsAPI

yukiisbored:
Hi @mariobehling, Which GCI task can I do for this?

Orbiter:
disabled in https://github.com/loklak/loklak_server/commit/9bb1a21a6d0acff92659f0e130c84b61fd6b9c9b

This was the main cause that api.loklak.org was unavailable in the last three months.
Every time I had to clean up the mess that the storage process had caused. Fortunately I had a back-up of the settings file. The stored file was a big mess, full of binary data and caused that loklak was then unable to start again.

mariobehling:
@Orbiter We need a proper handling, maybe difference between config and settings file? How can we get this feature?

PR #926, opened 2017-01-04
Fix Query for POST requests

yukiisbored:
Instead of replacing qm when it's a POST request, it's better to
add the results rather than replacing all of the contents.

qm now contains the content of the request parameters and to reduce
code duplication, it's placed on the constructor.

initGET has been modified to complement the modification of the
constructor.

initPOST is deprecated because it's purpose has been replaced by the
constructor.

Fixes https://github.com/loklak/loklak_server/issues/913

### Short description
I have:

### For the reviewers
I have:

### DIFF
--- a/src/org/loklak/server/Query.java
+++ b/src/org/loklak/server/Query.java
@@ -24,6 +24,7 @@ import java.text.ParseException;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Enumeration;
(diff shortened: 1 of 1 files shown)


yukiisbored:
This PR is required to make https://github.com/loklak/loklak_server/issues/919 possible, because updating settings are done via POST http requests.

mariobehling:
Thank you!

PR #955, opened 2017-01-07
Fixes #922: Added docs/api.md

### LINKED ISSUE #922: Move http://loklak.org/api.html to docs/API
As part of the process to unify documentation, please move http://loklak.org/api.html to docs/API.

fatimarafiqui:
Fixed #922 
@mariobehling @jig08

### DIFF
--- a/docs/download.md
+++ b/docs/download.md
@@ -38,7 +38,10 @@ To import a message dump (which you get from the [dump directory](http://loklak.
              ⌊imported                               // processed dump files from the import folder are moved here
              ⌊own                                    // dump files which this application creates, accessible at /dump/
 ```
-Imported dumps are not deleted, but moved to the `imported` directory. Because extracted hashtags, links and user names are not part of the dump, this is done during the import process and written to the elasticsearch index. While imports are running, you can use the [/api/status.json](http://loklak.org/api.html#status) servlet to moniotor the import progress.
+Imported dumps are not deleted, but moved to the `imported` directory. Because extracted hashtags, links and user names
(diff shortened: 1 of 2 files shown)


mariobehling:
@fatimarafiqui Thank you. Unfortunately the new developer site is not online yet. Therefore, please keep the html in its place for now as well. Only copy the api.md.

Please keep the original in the html until we have the developer site re-instated. Thanks!

docs/download.md
```diff
@@ -38,7 +38,10 @@ To import a message dump (which you get from the [dump directory](http://loklak.
              ⌊imported                               // processed dump files from the import folder are moved here
              ⌊own                                    // dump files which this application creates, accessible at /dump/
 ```
-Imported dumps are not deleted, but moved to the `imported` directory. Because extracted hashtags, links and user names are not part of the dump, this is done during the import process and written to the elasticsearch index. While imports are running, you can use the [/api/status.json](http://loklak.org/api.html#status) servlet to moniotor the import progress.
+Imported dumps are not deleted, but moved to the `imported` directory. Because extracted hashtags, links and user names
+are not part of the dump, this is done during the import process and written to the elasticsearch index. While imports
+are running, you can use the
+[/api/status.json](https://github.com/loklak/loklak_server/blob/development/docs/api.md) servlet to moniotor the import progress.
```
daminisatya: Ok before that let's fix this typo too. Moniotor -> monitor

fatimarafiqui:
@mariobehling @daminisatya Fixed as suggested.

mariobehling:
Thanks!

Issue #965, opened 2017-01-10
Add support for non-daemonize on start scripts


yukiisbored:
Currently, the start scripts only starts Loklak daemonized, this is bad for Docker, Heroku, and other container services because they don't know when Loklak is down, crashed, etc.

PR #966, opened 2017-01-10
start: Add support for non-daemonize option

yukiisbored:
The start script has been modified to launch Loklak without forking to
the background. The other scripts that are used for services like
Dokcer, Heroku, and Bluemix are modified to launch Loklak without
daemonization making them know the actual program and aware of it's
actual process.

Closes https://github.com/loklak/loklak_server/issues/965

### Short description
I have:

### For the reviewers
I have:
- [x] Reviewed this pull request by an authorized contributor.
- [x] The reviewer is assigned to the pull request.

### DIFF
--- a/bin/start.sh
+++ b/bin/start.sh
@@ -8,7 +8,7 @@ cd $(dirname $0)/..
 # Execute preload script
 source bin/.preload.sh
 
-while getopts ":Id" opt; do
+while getopts ":Idn" opt; do
     case $opt in
         I)
             SKIP_INSTALL_CHECK=1
@@ -16,10 +16,14 @@ while getopts ":Id" opt; do
         d)
             SKIP_WAITING=1
             ;;
+        n)
+            DO_NOT_DAEMONIZE=1
+            ;;
         \?)
             echo "Usage: $0 [options...]"
             echo -e " -I\tIgnore installation config"
             echo -e " -d\tSkip waiting for Loklak"
+            echo -e " -n\tDo not Daemonize"
             exit 1
             ;;
     esac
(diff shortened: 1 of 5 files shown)


bin/start.sh
```diff
@@ -58,10 +62,23 @@ if [ ! -f $INSTALLATIONCONFIG ] && [[ $SKIP_INSTALL_CHECK -eq 0 ]]; then
...
+
 echo "starting loklak"
 echo "startup" > $STARTUPFILE
 
-cmdline="$cmdline -server -classpath $CLASSPATH -Dlog4j.configurationFile=$LOGCONFIG org.loklak.LoklakServer >> data/loklak.log 2>&1 &";
+cmdline="$cmdline -server -classpath $CLASSPATH -Dlog4j.configurationFile=$LOGCONFIG org.loklak.LoklakServer";
+
+# If DO_NOT_DAEMONIZE, pass it to the java command, end of this script.
+if [[ $DO_NOT_DAEMONIZE -eq 1 ]]; then
+    exec $cmdline
```
niccokunzmann: Can I assume that this means the non deamonized version only prints to stdout and not to a log file?

bin/start.sh
```diff
@@ -58,10 +62,23 @@ if [ ! -f $INSTALLATIONCONFIG ] && [[ $SKIP_INSTALL_CHECK -eq 0 ]]; then
...
+fi
+
+cmdline="$cmdline >> data/loklak.log 2>&1 &"
 
 eval $cmdline
```
niccokunzmann: To me it looks like the cmdline is executed twice when not deamon. Could that be?

bin/start.sh
yukiisbored: Yes, See
https://github.com/loklak/loklak_server/pull/966/files#diff-97aedc598f09a22e269a16d91e204584R68

bin/start.sh
yukiisbored: @niccokunzmann No, it adds `>> data/loklak.log 2>&1 &" which redirects `stdout` and `stderr` in case if loklak crashed before log4j2 and the `&` symbol at the end is what actually makes it run daemonized

yukiisbored: If the command in `eval` is done the script will continue, unlike `exec` it'll pass the execution to it and the script stops on that line.

mariobehling:
Thank you!

Issue #1013, opened 2017-01-18
Ensure _assets folder are copied to gh-pages from /docs on main branches


mariobehling:
Images for the Readme.rst were added to the folder at /docs/_assets. Please ensure the folder gest copied to the gh-pages branch and the links to the images are updated when the Readme.rst is copied into the gh-pages branch to generate the documentation site (http://dev.loklak.org/).

![screenshot from 2017-01-18 20-51-16](https://cloud.githubusercontent.com/assets/1583873/22080659/adf75268-ddc0-11e6-8381-f3250670a440.png)

PR #1017, opened 2017-01-19
README: Fix missing pictures on Docs

yukiisbored:
Closes https://github.com/loklak/loklak_server/issues/1013

I have:

### For the reviewers
I have:


mariobehling:
@yukiisbored I understand why you are doing this, but external services, that might track users and which are not associated with us to include images are really not desired. While this solution "works" in the sense that it displays the images, it does not work in the sense, that we are further implementing additional services, that we do not have any relation to.

Please help to find a solution that ensures all data is kept on GitHub.

@jig08 Please help to avoid to use services that we do not know. A big issue was also that recently on the GCI website there was a service included that spammed people with advertising. External services of unknown providers can increase the risk of spam, advertising and security issues. We should avoid them as much as possible.

mariobehling:
Unfortunately I have to revert this PR therefore. Such a pity.

jig08:
>@jig08 Please help to avoid to use services that we do not know. A big issue was also that recently on the GCI website there was a service included that spammed people with advertising. External services of unknown providers can increase the risk of spam, advertising and security issues. We should avoid them as much as possible.

Advertising & Spamming via GCI website is pitiful indeed, it shall be taken care in future.

PR #1062, opened 2017-03-17
Fixes #1060: Increase default Xmx value

### LINKED ISSUE #1060: Increase default Xmx value
Continuation of https://github.com/loklak/loklak_server/issues/1049.

Due to "first timers" problems regarding Loklak running out of memory is due to the fact the default allocated memory for Java is set to 600 MB by default (defined here https://github.com/loklak/loklak_server/blob/development/conf/config.properties#L102).

I think it should be increased because Loklak requires more than 600 MB and 600 MB is not enough to run Loklak for a day or two.

vibhcool:
This PR increases the default Xmx value from 600mb to 1024mb

### DIFF
--- a/bin/.preload.sh
+++ b/bin/.preload.sh
@@ -7,7 +7,7 @@ DFAULTCONFIG="conf/config.properties"
 CUSTOMCONFIG="data/settings/customized_config.properties"
 LOGCONFIG="conf/logs/log-to-file.properties"
 STARTUPFILE="data/startup.tmp"
-DFAULTXmx="-Xmx800m";
+DFAULTXmx="1024m";
 CUSTOMXmx=""
 
 mkdir -p data/settings
--- a/conf/config.properties
+++ b/conf/config.properties
@@ -99,7 +99,7 @@ search.timeout = 3000
 server.blacklist=
 
 # memory allocation
-Xmx=600m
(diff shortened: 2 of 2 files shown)


vibhcool:
open for review :)

bin/.preload.sh
```diff
@@ -7,7 +7,7 @@ DFAULTCONFIG="conf/config.properties"
 CUSTOMCONFIG="data/settings/customized_config.properties"
 LOGCONFIG="conf/logs/log-to-file.properties"
 STARTUPFILE="data/startup.tmp"
-DFAULTXmx="-Xmx800m";
+DFAULTXmx="1024m";
```
yukiisbored: `DFAULTXmx` is one of the command line arguments for `java` to run Loklak. It's not just a value like what's on `config.properties`.

vibhcool: @yukiisbored DFAULTXmx is concatinating with string "-Xmx" to become command line argument of java. here -> https://github.com/loklak/loklak_server/blob/development/bin/.preload.sh#L48

yukiisbored: The variable is used if the default and customized configuration file doesn't exist. Also, yeah you're correct that it just have to state the size without the argument.

Issue #1051, opened 2017-03-14
Docker: Add volume path


niccokunzmann:
- Issue type: Bug report/Feature request 

### Short description

Currently, the docker image does not have a volume nor do i know where the data is stored that should be available when a new loklak version appears

### Expected behaviour

I would like to have a volume in

- [ ] Dockerfile
- [ ] docker-compose.yml

### Actual behaviour

There is no volume. I suppose, once the image is updated, all data is lost.

### Usecase

With the volume, I can update the docke rimage to the latest version automatically
### Description

There should be a volume for every folder that should be kept from image to image.
Steps to resolve:

- [X] read about docker volumes if they are updated to later versions  
  According to the documentation "Changes to a data volume will not be included when you update an image." - [source](https://docs.docker.com/engine/tutorials/dockervolumes/#data-volumes)
- [X] find all folders with data  
  The folder is "data" which translates to "/app/data" in the container
- add the folders as a volume
  - [ ] to Dockerfile 
    This allows persistence and saving the data when running one image
  - [ ] to docker-compose.yml
    This allows storing the data, when the image is updated to the latest version

yukiisbored:
Every runtime generated data is always stored on `data/`.

itspritish:
I's looking for a first-timer-only and found few, but i thought this one i could give a try as i have some hands on with docker. I could try if i get some help to reproduce this issue on my Ubuntu.

niccokunzmann:
Sure, please go ahead. If you have any questions, you can ask!

itspritish:
Ran loklak server as standalone and as a docker container. The folder that i see getting created/ updated are mostly in /loklak_server/data/ like index, dump folders. Also the data/loklak.log file grew pretty heavily. So is this the folder that need to be mapped to volume? 
If so this can be added in Dockerfile/ docker-compose.yml. BTW for some reason it doesn't allow me to just use /data as volume mapped folder. No folder gets created inside the container upon start. So I'm forced to add /loklak_server/data.

niccokunzmann:
> are more?

yukiisbored suggested that folder. I assume it is the only one.

itspritish:
Observed something (issue?) when i use docker-compose build/ up. It does not generate loklak.log file in loklak_server/data folder, it rather starts printing the log on the command console. I also tried docker-compose up -d option, no logs, neither in /data nor on the console. I initially thought it might be related to my changes, but its not. 
[...]

PR #1089, opened 2017-03-31
Creating Volume for persistence while deploying via docker, Issue #1051

itspritish:
Made changes to Dockerfile, also updated docker-compose.yml to version: '2' so that i can use volume naming. The volume name would be ImageName_VolumeName. If not named, docker will default volume name to long hash string. I have given volume name as 'data' and mapped it to existing folder 'loklak_server/data'. docker volume inspect <vol_name> will show the folder where  data is stored on the host machine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ RUN apk update && apk add openjdk8 git bash && \
 WORKDIR /loklak_server
 
 # Create Volumes
-VOLUME ["/loklak_server/data","/loklak_server/backup",]
+VOLUME ["/loklak_server/data","/loklak_server/backup"]
 
 # start loklak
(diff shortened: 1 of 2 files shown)


niccokunzmann:
Thanks! I believe this is excatly what was needed!

If you reference the issue like this: #1051 It turns up as a discussion.
As a style guide, we like squashing commits (I can do this) and naming them "fix #1051", so it closes the issue automatically. I will see if I can do this, too.

niccokunzmann:
Looks like it worked: https://github.com/loklak/loklak_server/commits/development

itspritish:
Thanks a ton for helping me with my first :) will look forward for more !

PR #1098, opened 2017-04-09
Add Unit Tests to Loklak Server

vibhcool:
Presently Loklak Server has no Automated Unit Testing 

Pre-Settings:

I have added a profile @loklak_test to twitter for testing of TwitterScraper.java

Changes : 
1) commit: Add unit tests for TwitterScraper.java
 - creates unit test for TwitterScraper. (more tests needed to be added, working on it)
 - creates TestRunner which can run all tests added to it.
 - modifies TwitterScraper.java so that it can be tested. i.e. change all private access modifier to public access modifier.

2) commit: Add data file to test JSONRandomAccessFileTest.java
 - fixes existing test by adding datafile to be tested

3) commit: Set up unit tests build in loklak Server
 - fixes:  issue #883
 - add test building settings in build.gradle
 - create test.sh to run tests (TestRunner.java)

Steps to run unit tests :-

Step 1 : ./gradlew test
Step 2 : bin/test.sh

PR #1106, opened 2017-04-17
More readable version of assert checking and another one unit test, fix bug.

vrnsky:
This pull request for #1098 
### Short description
I make next changes:
1. Change ``` assertEquals ``` to the ``` assertThat```, second version is more readable, because it provide more detail information when test fails.
2. I started write unit test for ```YoutubeScraper.java``` but it fail on next numbers ```4 098 300```, regular expression does not remove white spaces. I fix this by using CharMatcher from Guava. On this line https://github.com/loklak/loklak_server/blob/development/src/org/loklak/harvester/YoutubeScraper.java#L208
3. I see ```TwitterScraperTest.java``` and noticed that for Twitter exist Timeline which purpose is presentation.
I don't see something similar four youtube videos.


mariobehling:
@vrnsky Thanks a lot. Could you help us to keep everyone involved and put up an issue about this as well. Then match the issue by mentioning it in the pull request. 

I recently wrote a blog post how we usually work. Would be great if you could help to make it easier for reviewers as well.

http://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia/

vrnsky:
@mariobehling I read blog post, and add number of issue, also add link to the line on which YoutubeScrapper fail. What information still required?

vibhcool:
@vrnsky nice pr :smile:  
but I think as a best practice, this pr shall be split to youtubescraper pr and and twitterscrapertest pr

And when working with git, you shall create new branch, work on it and pull request from that branch. This keeps you out of any trouble you may face. And keep development branch untouched.

vibhcool:
@vrnsky and one more thing, bin/test.sh runs harvester tests, could you add new youtubetest class in TestRunner.java :-)

vrnsky:
@vibhcool that you for your suggestion at the next pull requests I will always follow your advice:)

mariobehling:
@vrnsky Thank you. You referenced a pull request to a pull request. The idea is, that other devs understand what is going on. So, the flow is 1. open an issue and describe the bug or feature and 2. provide a pull request with the solution. Could you help us with that, please?

src/org/loklak/harvester/YoutubeScraper.java
```diff
@@ -205,7 +206,8 @@ public static JSONObject parseVideo(final BufferedReader br) throws IOException
     }
     
     private static long parseNumber(String n) throws NumberFormatException {
-        return Long.parseLong(numberfix.matcher(n).replaceAll(""));
+        String withoutSpaces = CharMatcher.WHITESPACE.removeFrom(n);
```
singhpratyush: Please explain this change. How will it tackle `,`?

test/org/loklak/harvester/TwitterScraperTest.java
```diff
@@ -33,7 +31,7 @@ public void testPrepareSearchURL() {
             url = TwitterScraper.prepareSearchURL(query[i]);
 
             //compare urls with urls created
-            assertEquals(url, out_url[i]);
+            assertThat(out_url[i], is(url));
```
singhpratyush: Should it not be `assertThat(url, is(equalTo(out_url[i])));`, as the convention is `assertThat(actual, is(equalTo(expected)));`

test/org/loklak/harvester/TwitterScraperTest.java
```diff
@@ -65,20 +63,24 @@ public void testSearch() {
...
-            if (tweet_list == null) tweet_list = new Timeline[]{new Timeline(order), new Timeline(order)};
+            if (tweet_list == null) {
+                tweet_list = new Timeline[]{new Timeline(order), new Timeline(order)};
+            }
         }
 
         ftweet_list = processTweetList(tweet_list);
         //compare no. of tweets with fetched no. of tweets
-        assertEquals(2, ftweet_list.size());
+        assertThat(ftweet_list.size(), is(2));
```
singhpratyush: I think this should be
```java
assertThat(ftweet_list.size(), is(equalTo(2)));
```

src/org/loklak/harvester/YoutubeScraper.java
vrnsky: YoutubeScraper fails with NumberFormatException on number which contains an whitespaces. Long.parseLong expect that you give it number without whitespaces. Current regexp does not remove whitespaces from string and I am decided remove spaces by using CharMatcher from Guava

test/org/loklak/harvester/TwitterScraperTest.java
vrnsky: http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html
Inside is matcher calls equals of the string. I think that equalTo is redudant for this case and do not this assert is more readable

test/org/loklak/harvester/TwitterScraperTest.java
vrnsky: At this assert using equalTo is not needs because assert compare primitive type

src/org/loklak/harvester/YoutubeScraper.java
singhpratyush: But this implementation will break because comma in numbers will not be considered. A better solution would be to include space in the regex used for replacement.

test/org/loklak/harvester/TwitterScraperTest.java
singhpratyush: But we must keep up with the conventions. The expected output should go in as the second argument to `assertThat`.

Orbiter:
merged this even if it breaks the YoutubeScraper for now. Added an issue for that:
https://github.com/loklak/loklak_server/issues/1143

@vrnsky Thank you for the pull request. Please take care for the issue above!

PR #1147, opened 2017-05-10
fixes #1146: Check for TwitterFactory before getting instance

### LINKED ISSUE #1146: loklak fails to start with Kaizen harvester
- Issue type: Bug report 

### Short description

loklak fails to instantiate `KaizenHarvester` - 
```java
Exception in thread "main" java.lang.NullPointerException
	at org.loklak.harvester.strategy.KaizenHarvester.<init>(KaizenHarvester.java:64)
	at org.loklak.LoklakServer.initializeHarvester(LoklakServer.java:406)
	at org.loklak.LoklakServer.main(LoklakServer.java:294)
```

`KaizenHarvester.java:64` -
```java
        twitter = TwitterAPI.getAppTwitterFactory().getInstance();
```

**Reason**: `TwitterAPI.getAppTwitterFactory()` returns `null` because of no API keys in `config.properties`.

### Environment

 - Operating system: CentOS 7
 - Software version: Even with `development`
 - Software source: Github Repository

### Steps to reproduce

1. Change `harvester.type` to `kaizen` in `config.properties`.
2. Make sure that there are no API keys.
3. See the log.

### Expected behaviour

There should be no error and loklak should start smoothly because Kaizen can work without any API keys too.

### Actual behaviour

The startup fails with mentioned stack trace.

singhpratyush:
### Short description
Fixes #1146.

Included check for `twitterFactory` not being null before getting instance to avoid `NullPointerException`.
I have:

### For the reviewers
I have:

### DIFF
--- a/src/org/loklak/harvester/strategy/KaizenHarvester.java
+++ b/src/org/loklak/harvester/strategy/KaizenHarvester.java
@@ -14,6 +14,7 @@ import twitter4j.Location;
 import twitter4j.Trend;
 import twitter4j.Twitter;
 import twitter4j.TwitterException;
+import twitter4j.TwitterFactory;
 
 import java.io.IOException;
 import java.text.DateFormat;
@@ -60,7 +61,10 @@ public class KaizenHarvester implements Harvester {
 
         random = new Random();
 
-        twitter = TwitterAPI.getAppTwitterFactory().getInstance();
+        TwitterFactory twitterFactory = TwitterAPI.getAppTwitterFactory();
+
+        if (twitterFactory != null)
+            twitter = twitterFactory.getInstance();
 
         if (twitter == null)
             DAO.log("Kaizen can utilize Twitter API to get more queries, If you want to use it, " +


vibhcool:
@singhpratyush please rebase the commit

src/org/loklak/harvester/strategy/KaizenHarvester.java
```diff
@@ -61,7 +62,10 @@ public KaizenHarvester() {
 
         random = new Random();
 
-        twitter = TwitterAPI.getAppTwitterFactory().getInstance();
+        TwitterFactory twitterFactory = TwitterAPI.getAppTwitterFactory();
+
+        if (twitterFactory != null)
+            twitter = twitterFactory.getInstance();
 
         if (twitter == null)
```
vibhcool: you may join this as else condition with if condition in line 67

singhpratyush: Though both the cases will result in same outcomes, the first condition is to check if the `twitterFactory` is `null`. While the second is to give out a log if `twitter` is `null`, and hence is not the * else * part of the first.

singhpratyush:
@SarishiNoHara: That was a really good link. Thanks for sharing.

But I am afraid that this isn't applicable in this case as `TwitterFactory` is `final` class.
```java
private static TwitterFactory nullTwitterFactory = new TwitterFactory() {
    @Override
    public Twitter getInstance() {
        return null;
    }
};
```
```java
loklak_server/src/org/loklak/harvester/TwitterAPI.java:153: error: cannot inherit from final TwitterFactory
    private static TwitterFactory nullTwitterFactory = new TwitterFactory() {
                                                           ^
```

What would you suggest in this situation?

SarishiNoHara:
Nothing wrong with this code, but it will be much more complex if we check long chain of method calls for `NullPointerException`.

Yesterday I said using try catch block, but now I think it is inconvenient to locate `NullPointerException` if the object has a chain of method calls.

Here is a [way in Java 8](http://www.oracle.com/technetwork/articles/java/java8-optional-2175753.html) to handle null object, but anyway your code is reasonable and should be approved.

By the way, I don't quite understand what the wrapper class means, is that a class for exceptions?

singhpratyush:
The wrapper class would have been something that would work as proxy to `TwitterFactory` class, where we could have defined it in such a way that it would return `null` if we call `getInstance()` on a **semantically** `null` object.

PR #1133, opened 2017-05-07
Fixes part of #1132: Add unit test to check TwitterScraper output

vibhcool:
This PR is related to issue #1132

This PR:-
 - Adds Unit Test to compare TwitterScraper.search() method's output with the expected tweet data
 - Changes the access specifiers in TwitterTweet class and MessageEntry class so that they can be tested.
 - Since the output of images during scraping of tweets isn't correct. It is not tested and added in TODO.

PR #1151, opened 2017-05-11
Fix related to #1154: fix access specifiers

### LINKED ISSUE #1154: Using of getter methods instead of modifying access specifiers
In PR #1133 , access modifiers were modified instead of using getter methods. It is a bad approach of accessing the variables. This shall be fixed.

vibhcool:
This PR fixes:-
(Fixes #1154)
1) Replace back attributes of TwitterTweet object with getter methods present.
2) Restore access specifiers of the attributes modified


mariobehling:
Which issue is this related to? You are referencing another PR. Please match it with the issue. So we understand the background of this better. What is the problem you are solving?

vibhcool:
Added the issue, please review :)

src/org/loklak/harvester/TwitterScraper.java
```diff
@@ -416,8 +416,8 @@ public String toString() {
     
     public static class TwitterTweet extends MessageEntry implements Runnable {
 
-        public final Semaphore ready;
-        public UserEntry user;
+        private final Semaphore ready;
+        private UserEntry user;
         public boolean writeToIndex, writeToBackend;
```
kavithaenair: Is this one private or public?As i can see you have modified this variable in #1133 . And can you please use one line declarations as we started solving codacy issues, don't want to add this up to the issue there again.

test/org/loklak/harvester/TwitterScraperTest.java
```diff
@@ -107,16 +107,20 @@ public void testSearch() {
...
+        String created_date = dateFormatChange(String.valueOf(tweet.getCreatedAt()));
         assertTrue(created_date.equals(tweet_check.get("created_at")));
         
-        assertTrue(String.valueOf(tweet.screen_name).equals(tweet_check.get("screen_name")));
-        assertTrue(String.valueOf(tweet.source_type).equals(tweet_check.get("source_type")));
-        assertTrue(String.valueOf(tweet.provider_type).equals(tweet_check.get("provider_type")));
-        assertTrue(String.valueOf(tweet.text).equals(tweet_check.get("text")));
+        assertTrue(String.valueOf(tweet.getScreenName()).equals(tweet_check.get("screen_name")));
+        assertTrue(String.valueOf(tweet.getSourceType()).equals(tweet_check.get("source_type")));
+        assertTrue(String.valueOf(tweet.getProviderType()).equals(tweet_check.get("provider_type")));
```
SKrPl: Please use `assertEquals` when you are checking equality of two objects, it looks more intuitive. And `assertEquals` uses `equals` method internally.

src/org/loklak/harvester/TwitterScraper.java
```diff
@@ -416,9 +416,10 @@ public String toString() {
     
     public static class TwitterTweet extends MessageEntry implements Runnable {
 
-        public final Semaphore ready;
-        public UserEntry user;
-        public boolean writeToIndex, writeToBackend;
+        private final Semaphore ready;
+        private UserEntry user;
+        public boolean writeToIndex;
+        public boolean writeToBackend;
```
vibhcool: @kavithaenair fixed semantic error you pointed.
@kavithaenair @SKrPl Adding of getter methods only for unit-tests doesn't seem to be right. So I haven't dropped changes here. what do you say?

test/org/loklak/harvester/TwitterScraperTest.java
```diff
@@ -107,18 +110,24 @@ public void testSearch() {
...
-        assertTrue(tweet.writeToIndex);
-        assertTrue(tweet.writeToBackend);
+        assertEquals(tweet.getScreenName(), tweet_check.get("screen_name"));
+System.out.println(tweet.getSourceType()+ "   +++++   "+ tweet_check.get("source_type"));
+
+        assertEquals(String.valueOf(tweet.getSourceType()), tweet_check.get("source_type"));
+        assertEquals(String.valueOf(tweet.getProviderType()), tweet_check.get("provider_type"));
+        assertEquals(tweet.getText(), tweet_check.get("text"));
+        assertEquals(tweet.getPlaceId(), tweet_check.get("place_name"));
+        assertEquals(tweet.getPlaceName(), tweet_check.get("place_id"));
```
vibhcool: @SKrPl made the changes as you suggested.

src/org/loklak/harvester/TwitterScraper.java
vibhcool: @kavithaenair @SKrPl Adding of getter methods only for unit-tests doesn't seem to be right. So I haven't dropped changes here. what do you say?

kavithaenair: Adding getter methods looks more proper. Otherwise it looks fine to me. Good job 👍

SKrPl: @vibhcool Instead of using `getter` methods please use `Reflection`. For more please look into the SO link, [testing a class having private fields and methods](http://stackoverflow.com/questions/34571/how-do-i-test-a-class-that-has-private-methods-fields-or-inner-classes)

vibhcool: I will see to this :smile:

PR #1153, opened 2017-05-11
fix related to #1152: code refractoring for logging

vibhcool:
This PR:-
1) fixes a part of #1152 to refractor code to log data of some source files
2) adds debug log method in DAO.java

PR #1157, opened 2017-05-13
fix #1143: Fix NumberException in YoutubeScraper

### LINKED ISSUE #1143: YoutubeScraper fails with NumberFormatException
After merge of https://github.com/loklak/loklak_server/pull/1106/ the YoutubeScraper fails:

```
java.lang.NumberFormatException: For input string: "4,8Mio."
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Long.parseLong(Long.java:589)
	at java.lang.Long.parseLong(Long.java:631)
	at org.loklak.harvester.YoutubeScraper.parseNumber(YoutubeScraper.java:210)
	at org.loklak.harvester.YoutubeScraper.parseVideo(YoutubeScraper.java:137)
	at org.loklak.harvester.YoutubeScraperTest.whneTryParseVideoFromBufferedReaderShouldCheckThatIsGoodJsonObjectReturn(YoutubeScraperTest.java:40)
```

vibhcool:
fixes issue #1143

This is patch that considers the fact that a number in youtube can also be alphanumerical like 4.4M (4.4 million) or like a number like 123,123,345. So it considers just as a number

### DIFF
--- a/src/org/loklak/harvester/YoutubeScraper.java
+++ b/src/org/loklak/harvester/YoutubeScraper.java
@@ -36,6 +36,7 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.loklak.tools.CharacterCoding;
+import org.loklak.data.DAO;
 
 public class YoutubeScraper {
 
@@ -134,17 +135,17 @@ public class YoutubeScraper {
                 if ((p = input.indexOf("yt-subscriber-count")) >= 0) {
                     String subscriber_string = parseProp(input, p, "title");
                     if (subscriber_string == null) continue;
-                    json.put("youtube_subscriber", parseNumber(subscriber_string));
+                    json.put("youtube_subscriber", parseValue(subscriber_string));
                     continue;
                 }
                 if (input.indexOf("\"like this") > 0 && (p = input.indexOf("yt-uix-button-content")) >= 0) {
                     String likes_string = parseTag(input, p);
-                    json.put("youtube_likes", parseNumber(likes_string));
+                    json.put("youtube_likes", parseValue(likes_string));
                     continue;
                 }
                 if (input.indexOf("\"dislike this") > 0 && (p = input.indexOf("yt-uix-button-content")) >= 0) {
                     String dislikes_string = parseTag(input, p);
-                    json.put("youtube_dislikes", parseNumber(dislikes_string));
+                    json.put("youtube_dislikes", parseValue(dislikes_string));
                     continue;
                 }
                 if ((p = input.indexOf("watch-view-count")) >= 0) {
@@ -152,10 +153,7 @@ public class YoutubeScraper {
                     if (viewcount_string == null) continue;
                     viewcount_string = viewcount_string.replace(" views", "");
                     if (viewcount_string.length() == 0) continue;
-                    long viewcount = 0;
-                    // if there are no views, there may be a string saying "No". But this is done in all languages, so we just catch a NumberFormatException
-                    try {viewcount = parseNumber(viewcount_string);} catch (NumberFormatException e) {}
-                    json.put("youtube_viewcount", viewcount);
+                    json.put("youtube_viewcount", viewcount_string);
                     continue;
                 }
                 if ((p = input.indexOf("watch?v=")) >= 0) {
@@ -205,9 +203,14 @@ public class YoutubeScraper {
         return json;
     }
     
-    private static long parseNumber(String n) throws NumberFormatException {
-        String withoutSpaces = CharMatcher.WHITESPACE.removeFrom(n);
-        return Long.parseLong(withoutSpaces);
+    private static String parseValue(String n) {
+        String number_value = CharMatcher.WHITESPACE.removeFrom(n);
+        Pattern number_format = Pattern.compile("^(\\d{1,3})(((,\\d{3})*)|((\\.\\d+)?[MBK]))?$");
+        Matcher m = number_format.matcher(number_value);
+        if (!m.find()) {
+            DAO.severe("This is outputing wrong value. Need to be fixed");
+        }
+        return number_value;
     }
     
     private final static Pattern numberfix = Pattern.compile(",|\\.");


vibhcool:
I have little messed up...., This is not open for review

vibhcool:
Open for review :smile:

singhpratyush:
Does 4.4M need to be converted to 4,400,000?

kavithaenair:
@vibhcool can you please attach a screenshot of the final output? Try giving the input as mentioned in the issue #1143 .

vibhcool:
@singhpratyush 
the numbers on youtube we see can be of following formats:-
1) 123,123,312
2) 3
3) 4K meaning 4000
4) 5.3M .....etc
this fix checks them, get them and put them in JsonObject

vibhcool:
@kavithaenair to reproduce the error , Steps:-
1) comment this line -> https://github.com/loklak/loklak_server/blob/da1bb1089d29e2f079089968572340d4f567fafe/src/org/loklak/harvester/TwitterScraper.java#L188
(to prevent this flood the terminal, fixed this in PR #1153) 

2) run this command -> `./gradlew build`

3) run `bin/test.sh`
check the output at development branch with this

src/org/loklak/harvester/YoutubeScraper.java
```diff
@@ -205,9 +203,14 @@ public static JSONObject parseVideo(final BufferedReader br) throws IOException
         return json;
     }
     
-    private static long parseNumber(String n) throws NumberFormatException {
-        String withoutSpaces = CharMatcher.WHITESPACE.removeFrom(n);
-        return Long.parseLong(withoutSpaces);
+    private static String parseValue(String n) {
+        String number_value = CharMatcher.WHITESPACE.removeFrom(n);
+        Pattern number_format = Pattern.compile("((\\d)\\d{0,2})(,?\\d{3})*(\\.[\\d]+)?([MBK]?)");
```
singhpratyush: This may become `static` member of the class.

singhpratyush:
> this fix checks them, get them and put them in JsonObject

Is it not required here to convert the number to numeric-only value, like I mentioned in previous comment?

src/org/loklak/harvester/YoutubeScraper.java
vibhcool: this is static member of class, I didn't understand what you are suggesting :sweat_smile:

vibhcool:
youtube outputs some numbers like no. of subscriptions as 2M instead of it's exact value 2,353,367 , and so this 2M can be output like 2 million, 2M or 2000000 , but this numeric value isn't exact value and doesn't seem to be nice to as it is incorrect.

src/org/loklak/harvester/YoutubeScraper.java
singhpratyush: Since the variable `pattern_format` will be same every time, you can add it to the top of the class as a `private static` member.

src/org/loklak/harvester/YoutubeScraper.java
```diff
@@ -134,28 +135,25 @@ public static JSONObject parseVideo(final BufferedReader br) throws IOException
...
                     String subscriber_string = parseProp(input, p, "title");
                     if (subscriber_string == null) continue;
-                    json.put("youtube_subscriber", parseNumber(subscriber_string));
+                    json.put("youtube_subscriber", parseValue(subscriber_string));
                     continue;
                 }
                 if (input.indexOf("\"like this") > 0 && (p = input.indexOf("yt-uix-button-content")) >= 0) {
                     String likes_string = parseTag(input, p);
-                    json.put("youtube_likes", parseNumber(likes_string));
+                    json.put("youtube_likes", parseValue(likes_string));
```
Achint08: Hey @vibhcool . number of likes are always in numbers format . Only subscribers are shown in M,B,K format.  You can remove this.

src/org/loklak/harvester/YoutubeScraper.java
```diff
@@ -134,28 +135,25 @@ public static JSONObject parseVideo(final BufferedReader br) throws IOException
...
                     continue;
                 }
                 if (input.indexOf("\"dislike this") > 0 && (p = input.indexOf("yt-uix-button-content")) >= 0) {
                     String dislikes_string = parseTag(input, p);
-                    json.put("youtube_dislikes", parseNumber(dislikes_string));
+                    json.put("youtube_dislikes", parseValue(dislikes_string));
```
Achint08: Same goes for this.

src/org/loklak/harvester/YoutubeScraper.java
```diff
@@ -134,28 +135,25 @@ public static JSONObject parseVideo(final BufferedReader br) throws IOException
...
                 if ((p = input.indexOf("watch-view-count")) >= 0) {
                     String viewcount_string = parseTag(input, p);
                     if (viewcount_string == null) continue;
                     viewcount_string = viewcount_string.replace(" views", "");
                     if (viewcount_string.length() == 0) continue;
-                    long viewcount = 0;
-                    // if there are no views, there may be a string saying "No". But this is done in all languages, so we just catch a NumberFormatException
-                    try {viewcount = parseNumber(viewcount_string);} catch (NumberFormatException e) {}
-                    json.put("youtube_viewcount", viewcount);
+                    json.put("youtube_viewcount", viewcount_string);
```
Achint08: Same goes for this.

src/org/loklak/harvester/YoutubeScraper.java
singhpratyush: The regex will also accept values like `123,45` as `123` and `45`, which is not acceptable. Please fix it.

Also, the regex can be simplified. Please look into it.

Achint08: Regular expression is not working correctly for all the corner cases as well. Recheck it once.

vibhcool: @singhpratyush if we set it as private static member, it will consume memory for complete lifecycle of the YoutubeScraper oubject. It is used only in 1 method, So maybe local declaration suits it

vibhcool: @singhpratyush  123,45 isn't valid number, commas are put at the gap of 3 digits in numbers in internamtional system. 
Seeing for simplification :smile: 

@Achint08 could you share the cases where regex isn't working.

singhpratyush: > 123,45 isn't valid number

exactly, but it will get accepted by the regex as two discrete matches - `123` and `45`.

vibhcool: @singhpratyush the regex seems to be perfect :sweat_smile:

singhpratyush: @vibhcool: The regex you mentioned is 
```
((\d)\d{0,2})(,?\d{3})*(\.[\d]+)?([MBK]?)
```
If we try to parse `123,45`, here's what happens - 
```
  1 -> \d
 23 -> \d{2}
,45 -> No match, fall back, check for acceptance of previous substring
123 -> passes as \d\d{2}
  , -> No match, skip
  4 -> \d
  5 -> \d{1}
EOF -> 45 passes as \d\d{1}
```

So, we have two matches for `123,45`. What went wrong? The fact that your regex tries to match any possible substring. A better solution would be
```
^((\d)\d{0,2})(,?\d{3})*(\.[\d]+)?([MBK]?)$
```
This will do things fine, but regex can still be a lot simpler - 
```
^\d{1,3}(,\d{3})*(\.\d+)?[MBK]$
```
Breaking it down,
```
        ^ - Start of string
  \d{1,3} - X or XX or XXX
(,\d{3})* - ,XXX repeated zero or more times
 (\.\d+)? - .X or .XX or .XXX ...
   [MBK]? - M or B or K occurs 0 or 1 times
        $ - End of string
```

regex101.com is a goog website.

Achint08: @vibhcool  Hey!! Regular expression is not even giving match for 10M currently. 
As given by @singhpratyush  , `^\d{1,3}(,\d{3})*(\.\d+)?[MBK]?$` works fine. :)

Issue #1156, opened 2017-05-12
Produce-Consumer imbalance in KaizenHarvester


singhpratyush:
- Issue type: Bug report 

### Short description
The pace in which `KaizenHarvester` discovers new messages is much greater than the pace at which it harvests the queries. After some time, the queue size saturates to maximum and new messages are ignored.

This happens because the `KaizenHarvester` chooses between getting new suggestions and harvesting by tossing an unbiased coin (`random.nextBoolean()`) and it doesn't care about the saturation level of the queue.

### Suggested Solution
Instead of using `random.nextBoolean()`, we may compare `random.nextFloat()` with a value generated based on the size of queue filled vs. max allowable size, while the decision process will be same as current for an unrestricted queue.

@yukiisbored @daminisatya: Shall I proceed with this implementation?

PR #1158, opened 2017-05-13
Make harvesting decision biased for Kaizen

singhpratyush:
Solves part of #1156.

A probability is chosen as `queuries.size() / QUERIES_LIMIT`, which is compared to a randomly chosen target probability and decision is taken accordingly. In case of no limit on the queue size, probability to harvest is set to 0.5.

### Short description
I have:

### For the reviewers
I have:

### DIFF
--- a/src/org/loklak/harvester/strategy/KaizenHarvester.java
+++ b/src/org/loklak/harvester/strategy/KaizenHarvester.java
@@ -215,8 +215,14 @@ public class KaizenHarvester implements Harvester {
 
     @Override
     public int harvest() {
-        if (!queries.isEmpty() && random.nextBoolean())
+        float targetProb = random.nextFloat();
+        float prob = 0.5F;
+        if (QUERIES_LIMIT > 0) {
+            prob = queries.size() / (float)QUERIES_LIMIT;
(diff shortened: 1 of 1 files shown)


SarishiNoHara:
I think there might be a problem. For example, if `queries.size()` is 1, then `prob = queries.size() / (float)QUERIES_LIMIT` which is 0.002(default value of `QUERIES_LIMIT` is 500), but `targetProb` is uniformly distributed float value between 0.0 and 1.0, which is never smaller than `prob`, so the method never returns `harvestMessages()`.

singhpratyush:
@SarishiNoHara: That is the point. If the number of messages in the queue to harvest is less, it would be less likely for the harvester to go and execute the search query and it will go and grab more suggestions first.

But as the number of queries in queue increases, it becomes more probable that the harvester will harvest the queries instead of getting new ones.

SarishiNoHara:
@singhpratyush Oh yeah, you are right. I ignored that queries set will increase.

mariobehling:
@yukiisbored @Achint08 @vibhcool Please review, so we can progress.

Issue #1112, opened 2017-04-24
Additional constraints for images,videos not giving results restricted to media


Achint08:
- Issue type: Bug report 

### Short description

Searching for Query /image /audio /video /place , it is expected results to get restricted to tweets which have attached images, audio, video or place.  Searching for http://api.loklak.org/api/search.json?timezoneOffset =-330&q=fossasia+%2Fvideo doesn't give tweets restricted to videos. Same goes for /image.

vibhcool:
Please assign this issue to me. I want to work on this issue

Achint08:
@vibhcool You can start working on it asap. This issue is important so that we can proceed on loklak search with more features. Thank you. :)

daminisatya:
@vibhcool assigned!

singhpratyush:
@Achint08: You should be searching for \image and not /image.

Issue #1135, opened 2017-05-07
Output of image links is not correct


vibhcool:
- Issue type: Bug report

When we do api calls for '/api/search.json?q=query' , in image attribute (in json output), tweet post links also appear with image links.

Like for call loklak.org/api/search.json?q=fossasia , 

![screenshot from 2017-05-08 02-16-07](https://cloud.githubusercontent.com/assets/12577433/25785012/5d760508-3394-11e7-855a-585a4ea56bce.png)

djmgit:
I would like to work on this :)

singhpratyush:
[This](https://github.com/loklak/loklak_server/blob/development/src/org/loklak/harvester/TwitterScraper.java#L248) is where images are extracted. It uses `data-image-url` attr. Two image links are added means that there is an extra `data-image-url` that gets parsed.

For this, we need to check the HTML received while making the searches and make changes accordingly.

Another fix may involve restricting the image URLs that get added - if they do not match some pattern, they are rejected.

djmgit:
Thanks a lot for the pointers, I will send a PR soon :)

singhpratyush:
@djmgit: Make sure that @vibhcool is not working on it.

daminisatya:
@djmgit Assigned the issue

djmgit:
@singhpratyush I went through the code you referred. There as you mentioned image url is extracted using data-image-url attr then it is checked whether the image_url ends with jpg or png.If not it is reported as strange image link and also added to images set, and while extracting video thumbnail and animated gifs no check is done for gifs.
So  in case of image extraction I added an else clause where I added the image_url to images set only if it is not a invalid url. Here's a link to my fork where I did the change :  
https://github.com/djmgit/loklak_server/blob/imageoutput/src/org/loklak/harvester/TwitterScraper.java#L252

I also added a condition for checking gif in video thumbnail and animated gifs extraction.
Here is the change : 
https://github.com/djmgit/loklak_server/blob/imageoutput/src/org/loklak/harvester/TwitterScraper.java#L252

However a non image link is still getting added to the images set. Are images being added from somewhere else? Please do have a look at the changes which I made:
https://github.com/djmgit/loklak_server/blob/imageoutput/src/org/loklak/harvester/TwitterScraper.java
Thanks

vibhcool:
try links like this -> http://127.0.0.1:9000/api/search.json?q=video&source=twitter
that is , add `&source=twitter` when testing.
does this work??

[...]

PR #1164, opened 2017-05-16
Related to #1112:Add filter for images, videos

vibhcool:
This PR adds feature to :- 
1) searches for tweets with images if we set `filter=image` 
2) for tweets with videos if we set `filter=image`
3) and for both if we set `filter=image,video`

This PR:-

1) Partially fixes issue #1112 , doesn't work for cache(will push when this gets accepted, working on this)
2) outputs the total number of tweets with constraint(image or video) from total count of tweets. As loklak doesn't output more than 100 tweets due to performance issues.

Examples :- 
http://localhost:9000/api/search.json?q=video&filter=video,image&source=twitter
http://localhost:9000/api/search.json?q=paper&filter=image&source=twitter
http://localhost:9000/api/search.json?q=car&filter=video&source=twitter

### DIFF
--- a/src/org/loklak/harvester/TwitterScraper.java
+++ b/src/org/loklak/harvester/TwitterScraper.java
@@ -36,6 +36,8 @@ import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.HashSet;
+import java.util.Arrays;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Semaphore;
@@ -59,11 +61,13 @@ public class TwitterScraper {
 
     public static Timeline search(
             final String query,
+            final String filter,
             final Timeline.Order order,
             final boolean writeToIndex,
             final boolean writeToBackend,
             int jointime) {
-        Timeline[] tl = search(query, order, writeToIndex, writeToBackend);
+        
+        Timeline[] tl = search(query, filter.replaceAll("\\s",""), order, writeToIndex, writeToBackend);
         long timeout = System.currentTimeMillis() + jointime;
         for (MessageEntry me: tl[1]) {
             assert me instanceof TwitterTweet;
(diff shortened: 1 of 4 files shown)


src/org/loklak/harvester/TwitterScraper.java
```diff
@@ -290,7 +348,30 @@ public static String prepareSearchURL(final String query) {
...
+                    }
+                }
+
+            if (filter_array.contains("image") && images.size() < 1) {
+                props = new HashMap<String, prop>();
+                place_id = "";
+                place_name = "";
+                continue;
+            }
+
```
djmgit: @vibhcool please indent this if-block

src/org/loklak/harvester/TwitterScraper.java
```diff
@@ -165,24 +208,39 @@ public static String prepareSearchURL(final String query) {
...
         Timeline timelineWorking = new Timeline(order);
+        Set<String> filter_array = new HashSet<String>(Arrays.asList(
+                (filter.toLowerCase()).split(",")
+        ));
         String input;
         Map<String, prop> props = new HashMap<String, prop>();
         Set<String> images = new LinkedHashSet<>();
         Set<String> videos = new LinkedHashSet<>();
-        String place_id = "", place_name = "";
+        String place_id;
```
kavithaenair: This can be initialized here instead of doing it in line no 235 and 236. Line 235 is also not indented properly.

singhpratyush:
@vibhcool: Please write about how this patch works in the description.

src/org/loklak/harvester/TwitterScraper.java
kavithaenair: The image filter is showing emojis (images) if they are in the username.
For this:
http://localhost:9000/api/search.json?q=yolo&filter=image&source=twitter
the status link is:
https://twitter.com/xeooop/status/864815541197742081
the image link shown here is:
https://abs.twimg.com/emoji/v2/72x72/2744.png
But here the image shown is a character or emoji used in the username.
Is `filter=image` working fine?

vibhcool: this point about wrong images being shown is different issue related to #1135 i.e. storing wrong images.
filter=image just takes tweets which has images.

vibhcool:
@singhpratyush you are somewhat inferring opposite, but I got the point, that is about using better approach. will fix it . thanks 👍

vibhcool:
@djmgit , will add the test for this

singhpratyush:
@vibhcool: Could you please give the query for which it is showing non-image or non-video results?

Also, the issue you mentioned says that the search should be of type `github \image` and not `q=github&filter=image`. So you may directly check if those terms exist in the search query instead of introducing a new parameter.

vibhcool:
updated PR

1) Updated test for testing prepareSearchUrl method.
2) Some indentation and minor fixes in TwitterScraperTest.
3) Added Better approach for filtering tweets with videos.
 - I am directly fetching videos by setting `f=videos` in twitter url like `https://twitter.com/search?f=videos&vertical=default&q=paris&src=typd`
 - Not using this method for fetching images as on setting `f=images` , only images are fetched instead of fetching complete tweets having images.
4) Working on test for searching for tweets with filters.

vibhcool:
@daminisatya working on it

PR #1127, opened 2017-05-04
Auto-generate docs in dev.loklak.org repository

imujjwal96:
Related issue: #1109 
### Short description
I have:
- [x] There is a corresponding issue for this pull request.
- [x] Mentioned the Issue number in the pull request commit message `Fixes #1109 commit message`
- [x] There is only strictly only one commit per issue.

### For the reviewers
I have:
- [ ] Reviewed this pull request by an authorized contributor.
- [ ] The reviewer is assigned to the pull request.

Issue #1109, opened 2017-04-18
Instead of auto-generating the docs page in the gh-pages branch generate it in the dev.loklak.org repository


mariobehling:
Currently the documentation is generated a) in the gh-pages branch and b) transferred to the repository of dev.loklak.org.

Please change it as follows:
1. Do not generated the page in the gh-pages branch
2. Only add the page into the branch at dev.loklak.org

pri22296:
@mariobehling Since generation of docs is intended to be done only in the dev.loklak.org repository, will this mean that other repos would only have the rst/md files in the their docs directory? All sphinx setup in other repos such as this one would have to be removed?

mariobehling:
@pri22296 All other projects will have the same set up: a) Readme.rst and b) more documentation in their /docs folder and subfolders. The Readme.rst should be the start page of the subproject, e.g. the Readme.rst of the server project is the start page of dev.loklak.org/server/ and the loklak search project's Readme.rst is used to generate the start page of dev.loklak.org/search/.

singhpratyush:
@mariobehling: Are there any changes left to make for this issue or this can be closed?

PR #1195, opened 2017-05-27
Auto-generate docs in dev.loklak.org repository

imujjwal96:
Related issue: #1109

Following up https://github.com/loklak/loklak_server/pull/1127 which was broken due to mistakes while squashing and cherry picking commits.


singhpratyush:
@imujjwal96: Can you please describe the changes? Mostly for the deleted files, why are they no longer needed?

imujjwal96:
The documents are currently generated as
1. During Travis CI build, we use the sphinx framework to update and generate docs from the docs directory. 
2. The generated docs are then moved to gh-pages branch.
3. Correspondingly these generated docs are aggregated in the dev.loklak.org repository.

What is intended with this PR is that instead of generating and aggregating docs here, we directly move all the docs of various sub-projects of loklak to the dev.loklak.org repository and then use the sphinx framework to collectively build the docs.

The files that are deleted are the files that were needed for sphinx framework to build the docs. Since that is no longer intended, these files are no longer needed.

PR #1210, opened 2017-06-01
Add BlockingKaizenHarvester as harvesting strategy

### LINKED ISSUE #1189: Using blocking queue for KaizenHarvester
- Issue type: Question

### Short description
As of now, `KaizenHarvester` uses a simple queue in which queries to search for are stored. If there is a max size set for the queue, the harvester will not overshoot the number and cut off the new arriving queries when the queue is full.

A blocking queue would prevent threads from adding new queries to the queue until there is some room available.

The question: **Will this implementation will be advantageous over the current one?**

Related #1156.

singhpratyush:
### Short description

Fixes #1189.

BlockingKaizenHarvester uses a blocking queue so that queries produced are not dropped when the queue is full. Instead, the producers will be blocked for some time (`harvester.blocking_kaizen.block_time`) before they drop the queries.

I have:

### For the reviewers
I have:


src/org/loklak/harvester/strategy/BlockingKaizenHarvester.java
```diff
@@ -0,0 +1,61 @@
...
+            private int blockingTimeout = DAO.getConfig("harvester.blocking_kaizen.block_time", 120);
+
+            @Override
+            public boolean addQuery(String query) {
+                if (this.queries.contains(query)) {
+                    return false;
+                }
+                try {
+                    /*
+                    * We need to associate a timeout with the addition process as there are chances (though little)
```
hemantjadon: @singhpratyush How this timeout gonna avoid The Critical Section Problem, don't we need a proper Mutex here?

src/org/loklak/harvester/strategy/BlockingKaizenHarvester.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.loklak.data.DAO;
+
+public class BlockingKaizenHarvester extends KaizenHarvester {
+
+    /**
+     * This class uses blocking queue to store queries which need to be processed.
+     */
+    public BlockingKaizenHarvester() {
+        super(new KaizenQueries() {
```
yukiisbored: Please just make this into an inner class or something. This looks ugly and unmaintainable.

src/org/loklak/harvester/strategy/BlockingKaizenHarvester.java
yukiisbored: Using timers can result in a racing problem. Using Mutex is the proper solution.

singhpratyush: The [`BlockingQueue`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/BlockingQueue.html) handles all that internally.

Also, I didn't get that how timers can be a problem. The internal mutex will allow only one of the waiting queries will get added.

hemantjadon: Okay cool so blocking queue handles the Critical Section Problem, that's great :) 
But then why the timer is used here ?? I mean You have written in the comments specifi

singhpratyush: As I mentioned in the comment, there may be a case when all the harvester threads are trying to put in data at once and there are no consumers. In that case, one of them will get interrupted and drop the query.

singhpratyush:
I see a problem with such implementation in the harvester.

Every time a consumer takes a query from the task queue, it ends up producing a lot more. So, there will always be the case when everyone is trying to put things in the queue. And even after some thread gets interrupted, it will end up adding even more things to the queue.

@hemantjadon @yukiisbored: Is there a good way to do this? What I see is that since consumers end up producing even more tasks, we can't implement this in a memory efficient way without losing any queries.

hemantjadon:
@singhpratyush as you mentioned that every consumer ends up producing more number of entries and we can't drop or wait on any of them, this, in turn, becomes a complicated `CS` problem.
I don't know this even is in the paradigm of the standard producer-consumer problem. I don't know how this can be achieved. But one approach which I can think on top of my head is Use Multilevel queue scheme, and use first fulfill higher level queues and then lower level queues, and you start dropping queries when all the levels are full. I mean you cannot guarantee a correct solution but it might defer the eventual dropping of queries. We may do a further more research on this and try to find out the solution.

singhpratyush:
@hemantjadon: In the end, this will also be a matter of how much memory are we willing to give to the queues. If we can afford more memory for multiple queues, we can also afford a large single queue too.

But if by multilevel queue, you meant that a queue on multiple levels of storage, i.e. memory and disk, that can be something to think about. But I don't think that we are going to profit anything from it as the queue will be ever increasing (check for existence in the disk queue will be costly).

One possible solution can be that we send search queries as a suggestion to other nodes, who in turn can search for it. But I don't know if this would be good. @yukiisbored @sudheesh001: What are your views?

Another solution can be a priority queue, which contains more frequent terms at a higher level. And it has a fixed size so less frequent terms will be dropped. But I'm not sure about this one too as (I think) there will be a point when we are just dropping all the new queries that are coming to us.

hemantjadon:
@singhpratyush 
>But I'm not sure about this one too as (I think) there will be a point when we are just dropping all the new queries that are coming to us.

Yeah if we are fixing the size of the queue then it is not possible to serve all the queries, it's actually not possible.
What we have to make most of is how to make sure that essential entries don't get dropped. I don't know this is even possible or how well this can scale.

>One possible solution can be that we send search queries as a suggestion to other nodes, who in turn can search for it.

This seems to be some solution, but again don't know much about it. @sudheesh001 can shed some light on this.

>Multilevel queue

What I meant is prioritizing the entries in each queue Like if an entry comes and its already in some lower level queue it gets promoted. It's more like MFU algorithm but queued, MFU items get promoted to the top queue and the lower prority ones get down.

>But if by multilevel queue, you meant that a queue on multiple levels of storage, i.e. memory and disk, that can be something to think about.

No @singhpratyush this will be terribly slow to be handle sanely.

singhpratyush:
This needs more work from scratch before proceeding. Closing for now.

PR #1240, opened 2017-06-11
Add PriorityQueue harvesting strategy

### LINKED ISSUE #1238: Using priority queue for KaizenHarvester
We can replace the traditional queue with a priority queue in which the query with the highest priority will be the first to dequeue and the lower priority queries will be dropped when the queue is full. The higher the number of favourites and retweets, higher will be the priority of queries derived from that Tweet.

Related #1156 #1189.

singhpratyush:
### Short description

Fixes #1238.

Add score related to each Tweet based on retweet and favourite count, ranging from 0 to 1.
Formula used for calculation of score - 
```
a = retweet * 5 + favourite
score = a (a + 10exp(-0.01a))
```
[See Plot](http://fooplot.com/#W3sidHlwZSI6MCwiZXEiOiJ4Lyh4KzEwKmVeKC0wLjF4KSkiLCJjb2xvciI6IiMwMDAwMDAifSx7InR5cGUiOjEwMDAsIndpbmRvdyI6WyItMS42NjQyNTU2Mjc0ODUzOTYiLCIzMC40NjEzNjQxMTQxMDEwMyIsIi0zLjAwNTc0NDc2OTU3MzE4MDMiLCIyLjk0MDQxNDU5MzIxOTc2MzYiXX1d)

`LoklakServer.java` modified according to `.editorconfig`.

I have:

### For the reviewers
I have:

### DIFF
--- a/src/org/loklak/harvester/strategy/KaizenQueries.java
+++ b/src/org/loklak/harvester/strategy/KaizenQueries.java
@@ -7,7 +7,11 @@ import java.util.HashSet;
  */
 public abstract class KaizenQueries {
 
-    public abstract boolean addQuery(String query);
+    public boolean addQuery(String query) {
+        return this.addQuery(query, 0.5);
+    }
+
+    public abstract boolean addQuery(String query, double score);
 
     public abstract String getQuery();
 
@@ -26,7 +30,7 @@ public abstract class KaizenQueries {
             int queryLimit = qLimit;
 
             @Override
-            public boolean addQuery(String query) {
+            public boolean addQuery(String query, double score) {
                 if (this.queryLimit > 0 && this.queries.size() > this.queryLimit)
                     return false;
 
--- a/src/org/loklak/harvester/strategy/PriorityKaizenHarvester.java
+++ b/src/org/loklak/harvester/strategy/PriorityKaizenHarvester.java
@@ -1,0 +1,68 @@
+package org.loklak.harvester.strategy;
+
+import org.loklak.data.DAO;
+
+import java.util.Comparator;
+import java.util.PriorityQueue;
+import java.util.Queue;
+
+public class PriorityKaizenHarvester extends KaizenHarvester {
+
+    private static class PriorityKaizenQueries extends KaizenQueries {
+
+        private Comparator<ScoreWrapper> scoreComparator = (scoreWrapper, t1) -> (int) (scoreWrapper.score - t1.score);
+
+        private Queue<ScoreWrapper> queue;
+        private int maxSize;
+
+        private class ScoreWrapper {
+
+            private double score;
+            private String query;
+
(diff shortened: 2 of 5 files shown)


djmgit:
@singhpratyush please have a look at the codacy issues :)

singhpratyush:
@djmgit: Made the changes.

@hemantjadon @vibhcool @kavithaenair @SarishiNoHara @SKrPl @Achint08: Please review.

src/org/loklak/harvester/strategy/PriorityKaizenHarvester.java
```diff
@@ -0,0 +1,68 @@
...
+            ScoreWrapper(String m, double score) {
+                this.query = m;
+                this.score = score;
+            }
+
+        }
+
+        public PriorityKaizenQueries(int size) {
+            this.maxSize = size;
+            queue = new PriorityQueue<>(size, scoreComparator);
```
hemantjadon: Is this `PriorityQueue` thread safe like `BlockingQueue`?

singhpratyush: No, it is not. But we don't need to make it thread-safe I guess.

src/org/loklak/harvester/strategy/KaizenQueries.java
```diff
@@ -26,7 +30,7 @@ public static KaizenQueries getDefaultKaizenQueries(int qLimit) {
             int queryLimit = qLimit;
 
             @Override
-            public boolean addQuery(String query) {
+            public boolean addQuery(String query, double score) {
```
kavithaenair: Where is score variable used in this function? I don't understand the use of the variable.

src/org/loklak/harvester/strategy/PriorityKaizenHarvester.java
hemantjadon: Okay, so here the queries will get dropped only if we run out of memory, This seems good 👍

src/org/loklak/harvester/strategy/KaizenQueries.java
singhpratyush: It is passed, but not used in the Kaizen harvester.

kavithaenair: Why to pass then?

singhpratyush: Because for any type of `KaizenQueries`, the `private` methods of `KaizenHarvester` that signal the addition of new queries will be the same. i.e., it is not possible to modify the add call if the type of `KaizenQueries` is different. So, it is passed to all type of `KaizenQueries`' `addQuery` method, and it is up to them to use it.

PR #1299, opened 2017-07-05
Unshorten URL until applicable

### LINKED ISSUE #1297: Redirect unshortening service makes message collection much slower
- Issue type: Bug report

### Short description

Parent issue: #1296.

Due to making more number of requests before achieving a final, shortened URL, the time taken to collect a complete Tweet is very much. This, therefore, decreases the message collection rate by a significant factor.

It was introduced in #1122.

singhpratyush:
### Short description

Fixes #1297.

URL redirect unshortener would not keep unshortening URLs until it has reached some final URL. It will do so only until there are chances of URL being trackable.

I have:

### For the reviewers
I have:
- [x] Reviewed this pull request by an authorized contributor.
- [x] The reviewer is assigned to the pull request.

### DIFF
--- a/src/org/loklak/harvester/RedirectUnshortener.java
+++ b/src/org/loklak/harvester/RedirectUnshortener.java
@@ -6,12 +6,12 @@
  *  modify it under the terms of the GNU Lesser General Public
  *  License as published by the Free Software Foundation; either
  *  version 2.1 of the License, or (at your option) any later version.
- *  
+ *
  *  This library is distributed in the hope that it will be useful,
  *  but WITHOUT ANY WARRANTY; without even the implied warranty of
  *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  *  Lesser General Public License for more details.
- *  
+ *
  *  You should have received a copy of the GNU Lesser General Public License
  *  along with this program in the file lgpl21.txt
  *  If not, see <http://www.gnu.org/licenses/>.
@@ -57,7 +57,7 @@ public class RedirectUnshortener {
         "flip.it",
         "lnkd.in"
     };
-    
+
     private final static String[] untestedHosts = new String[] {
         "is.gd",
         "ta.gd",
@@ -72,15 +72,12 @@ public class RedirectUnshortener {
         "igg.me",
         "twiza.ru"
     };
-    
+
     public static String unShorten(String urlstring) {
         //long start = System.currentTimeMillis();
-        if (!isApplicable(urlstring)) {
-            return urlstring;
-        }
         try {
             int termination = 10; // loop for recursively shortened urls
-            while (termination-- > 0) {
+            while (termination-- > 0 && isApplicable(urlstring)) {
                 String unshortened = ClientConnection.getRedirect(urlstring);
                 if (unshortened.equals(urlstring)) {
                     return urlstring;
@@ -94,7 +91,7 @@ public class RedirectUnshortener {
             return urlstring;
         }
     }
-    
+
     private static boolean isApplicable(String urlstring) {
         String s = urlstring.toLowerCase();
         if (!s.startsWith("http://") && !s.startsWith("https://")) return false;
@@ -113,7 +110,7 @@ public class RedirectUnshortener {
         }
         return false;
     }
-    
+
     /**
      * this is the raw implementation if ClientConnection.getRedirect.
      * Surprisingly it's much slower, but some redirects cannot be discovered with the other
@@ -158,7 +155,7 @@ public class RedirectUnshortener {
         socket.close();
         return urlstring;
     }
-    
+
     public static void main(String[] args) {
         String[] test = new String[] {
                 "http://tmblr.co/Z6YPNx1jL1hHK",
@@ -193,11 +190,11 @@ public class RedirectUnshortener {
             }
         }
     }
-    
+
     /**
      * does not work:
      * https://tr.im/v31Rf
      * http://dlvr.it/8htd6W  // works on terminal but not here
      */
-    
+
 }
--- a/test/org/loklak/harvester/RedirectUnshortenerTest.java
+++ b/test/org/loklak/harvester/RedirectUnshortenerTest.java
@@ -21,7 +21,7 @@ public class RedirectUnshortenerTest {
         shortlinkMap.put("http://ow.ly/P9SX30ddSTI", "https://github.com/loklak/loklak_server/issues/1284");
         shortlinkMap.put("http://bit.do/blog-fossasia", "http://blog.fossasia.org/");
         shortlinkMap.put("http://fb.me/4lcXZsyyO", "https://www.facebook.com/permalink.php?story_fbid=1550262581900846&id=1381813572079082");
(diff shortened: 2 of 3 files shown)


singhpratyush:

src/org/loklak/harvester/RedirectUnshortener.java
```diff
@@ -6,12 +6,12 @@
  *  modify it under the terms of the GNU Lesser General Public
  *  License as published by the Free Software Foundation; either
  *  version 2.1 of the License, or (at your option) any later version.
- *  
+ *
```
sudheesh001: Unwanted change This is a standard template that gets prepended to the start of the file and needn't be modified.

src/org/loklak/harvester/RedirectUnshortener.java
```diff
@@ -6,12 +6,12 @@
...
  *  This library is distributed in the hope that it will be useful,
  *  but WITHOUT ANY WARRANTY; without even the implied warranty of
  *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  *  Lesser General Public License for more details.
- *  
+ *
```
sudheesh001: same here.

src/org/loklak/harvester/RedirectUnshortener.java
```diff
@@ -57,7 +57,7 @@
         "flip.it",
         "lnkd.in"
     };
```
sudheesh001: Is the string of working hosts ever going to get updated on a successful unshortening?

test/org/loklak/harvester/RedirectUnshortenerTest.java
```diff
@@ -21,7 +21,7 @@ public void testRedirectUnshortener() {
         shortlinkMap.put("http://ow.ly/P9SX30ddSTI", "https://github.com/loklak/loklak_server/issues/1284");
         shortlinkMap.put("http://bit.do/blog-fossasia", "http://blog.fossasia.org/");
         shortlinkMap.put("http://fb.me/4lcXZsyyO", "https://www.facebook.com/permalink.php?story_fbid=1550262581900846&id=1381813572079082");
-        shortlinkMap.put("http://wp.me/sf2B5-shorten", "https://en.blog.wordpress.com/2009/08/14/shorten/");
+        shortlinkMap.put("http://wp.me/sf2B5-shorten", "http://en.blog.wordpress.com/2009/08/14/shorten/");
```
sudheesh001: the shortened URL redirects to `https://`, Is there a reason why this was changed to `http`

src/org/loklak/harvester/RedirectUnshortener.java
singhpratyush: These changes are introduced by IDE due to [`.editorconfig`](https://github.com/loklak/loklak_server/blob/development/.editorconfig) file and are made to many other files too.

src/org/loklak/harvester/RedirectUnshortener.java
singhpratyush: No, they will not be updated.

test/org/loklak/harvester/RedirectUnshortenerTest.java
singhpratyush: Earlier (before this patch), the redirections were resolved until a final URL is reached. But it isn't required as the prime reason behind resolving redirects is to avoid tracking by other web services.

Once the URL is no longer a threat to anonymity, there is no need to resolve redirects any further as it will take time for no reason.

In this case, `http://en.blog...` is not a trackable URL so we'll not get redirected to `https://en.blog...`, so I changed it here.

src/org/loklak/harvester/RedirectUnshortener.java
sudheesh001: From the code I can see that it won't be, I am more over asking something to peek curiosity like, wouldn't that be a nicer thing.

test/org/loklak/harvester/RedirectUnshortenerTest.java
sudheesh001: No I don't seem to understand the statement, based on a URL shortened how can you say that it is a threat to anonymity or that it is a trackable or untrackable URL? if i have an URL which is http://b.ly vs https://b.ly where is the anonymity coming into place?

At what stage of URL redirect solution do you decide that the URL is not `anonymized`. Either way when your request hits the URL it will be tracked by who so ever does it. I don't see anonymity threats anywhere here.

My question is why does this test have `http://` instead of `https://` when `https://` works? Does the test fail when it's https?

singhpratyush: > My question is why does this test have `http://` instead of `https://` when `https://` works?

To explain this, let me take an example - 
`http://bit.ly/xyz` -> `http://something.new` -> `https://something.new`

Now, the URL unshortener would request `http://bit.ly/xyz` and get a `30X` response code asking for a redirect to `http://something.new`. Now, this URL is not `isApplicable()`, so the unshortener would not request this URL (and hence wouldn't get redirected to `https://something.new`). And when a user requests for `http://something.new` (after getting the URL from loklak's API), his browser redirects to `https://something.new`.

This would save loklak at least one HTTP request (maybe `https://something.new` -> `http://something.newer` -> `https://something.newer`) before finalizing a Tweet.

> Does the test fail when it's https?

After this change, yes.

> At what stage of URL redirect solution do you decide that the URL is not `anonymized`.

If it `isApplicable()`, it needs to be unshortened.


Just to mention, the [previous version of `RedirectUnshortener`](https://github.com/loklak/loklak_server/blob/ef587d75cf40c258e7996454c2a1838e599b618e/src/org/loklak/harvester/RedirectUnshortener.java#L80) also used this logic.

src/org/loklak/harvester/RedirectUnshortener.java
singhpratyush: I guess it would be better, but I don't see the two arrays having any impact on the code (or why are they two different arrays in the first place). So juggling a service from one array to other won't make a difference I guess.

test/org/loklak/harvester/RedirectUnshortenerTest.java
vibhcool: IMHO you could also keep both the present method to get the exact url and the applicable method and use this one.

singhpratyush: @hemantjadon: I never mentioned that `http` or `https` has anything to do with anonymity. I was just saying that we do not need to resolve a URL from `http` to `https` if it isn't from any url shortening service (which generally maintain stats about who is visiting the link, from where, interests, etc.).

singhpratyush:
I think that the purpose of this change needs to be clarified. This PR aims to reduce the number of HTTP requests that are required to be made before finalising a Tweet. And there are some redirects that we do not need to perform, hence, we can skip the URLs if they are not applicable for unshortening.

test/org/loklak/harvester/RedirectUnshortenerTest.java
hemantjadon: @singhpratyush Okay got it 👍

PR #1309, opened 2017-07-11
Add comparator for ScoreWrapper class and tests

### LINKED ISSUE #1308: Duplicate queries are added in queue for PriorityKaizen harvester
- Issue type: Bug report

### Short description
In `PriorityKaizenHarvester`, same queries with different scores are added to the queue. Ideally, same queries shouldn't exist twice in the queue.

singhpratyush:
Also fix precision issue with comparator

### Short description
Fixes #1308.

* Added comparator for `ScoreWrapper` objects.
* Fixed precision for comparison with the priority queue.
* Added tests.

I have:

### For the reviewers
I have:

### DIFF
--- a/src/org/loklak/harvester/strategy/PriorityKaizenHarvester.java
+++ b/src/org/loklak/harvester/strategy/PriorityKaizenHarvester.java
@@ -8,9 +8,9 @@ import java.util.Queue;
 
 public class PriorityKaizenHarvester extends KaizenHarvester {
 
-    private static class PriorityKaizenQueries extends KaizenQueries {
+    public static class PriorityKaizenQueries extends KaizenQueries {
 
-        private Comparator<ScoreWrapper> scoreComparator = (scoreWrapper, t1) -> (int) (scoreWrapper.score - t1.score);
+        private Comparator<ScoreWrapper> scoreComparator = (scoreWrapper, t1) -> (int) (1000 * (t1.score - scoreWrapper.score));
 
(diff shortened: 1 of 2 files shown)


singhpratyush:

src/org/loklak/harvester/strategy/PriorityKaizenHarvester.java
```diff
@@ -8,9 +8,9 @@
 
 public class PriorityKaizenHarvester extends KaizenHarvester {
 
-    private static class PriorityKaizenQueries extends KaizenQueries {
+    public static class PriorityKaizenQueries extends KaizenQueries {
 
-        private Comparator<ScoreWrapper> scoreComparator = (scoreWrapper, t1) -> (int) (scoreWrapper.score - t1.score);
+        private Comparator<ScoreWrapper> scoreComparator = (scoreWrapper, t1) -> (int) (1000 * (t1.score - scoreWrapper.score));
```
Achint08: Why multiplying by 1000?

singhpratyush: Since `score` is always between 0 and 1, the difference is always between -1 and 1. So when we cast it to `int`, the comparison fails for two numbers whose difference is less.

For example, if the difference is 0.2, they appear to be same as `(int)0.2 == 0`. But if we scale it by 1000, it becomes a reasonable number after casting and comparator would work properly.

PR #1319, opened 2017-07-14
Add API endpoint to serve video URLs

### LINKED ISSUE #1318: Add API endpoint to serve video URLs for various services
- Issue type: Feature request

### Short Description
The requirement is to add an endpoint that would serve video URLs for various services in loklak (whenever applicable), starting with Twitter.

Example request -
`http://loklak.org/api/videoUrlService.json?id=http://twitter.com/status/user1/status1&id=http://twitter.com/status/user2/status2`

Response - 
```json
{
    "metadata": {
        "ids": [
            "http://twitter.com/status/user1/status1",
            "http://twitter.com/status/user2/status2"
        ]
    },
    "videos": {
        "http://twitter.com/status/user1/status1": {
            "service": "Twitter",
            "status": 0,
            "message": "OK",
            "links": [
                "http://link.to/status/1_1.mp4",
                "http://link.to/status/1_2.mp4",            
            ]
        },
        "http://twitter.com/status/user2/status2": {
            "service": "Twitter",
            "status": 0,
            "message": "OK",
            "links": [
                "http://link.to/status/2_1.mp4",
                "http://link.to/status/2_2.mp4",            
            ]
        }
    }
}
```

singhpratyush:
### Short description

Fixes #1318.

Example: 
* http://35.184.98.133/api/videoUrlService.json?id=https://www.twitter.com/loklak_test/status/870536303569289216.

* http://35.184.98.133/api/videoUrlService.json?id=https://twitter.com/EXOGlobal/status/886182766970257409&id=https://twitter.com/KMbappe/status/885963850708865025

I have:

### For the reviewers
I have:


src/org/loklak/api/search/VideoUrlService.java
```diff
@@ -0,0 +1,122 @@
...
+import org.loklak.server.AbstractAPIHandler;
+import org.loklak.server.Authorization;
+import org.loklak.server.BaseUserRole;
+import org.loklak.server.Query;
+import org.loklak.tools.storage.JSONObjectWithDefault;
+
+
+public class VideoUrlService extends AbstractAPIHandler implements APIHandler {
+
+    private static final Pattern twitterUrlPattern = Pattern.compile("https?://(www\\.)?twitter.com(/.*?/status/[0-9]+)/?");
```
vibhcool: this variable can be defined in the method where it is used.

src/org/loklak/api/search/VideoUrlService.java
```diff
@@ -0,0 +1,122 @@
...
+    private static JSONObject handleNoMatchingService() {
+        JSONObject ret = new JSONObject(true);
+        ret.put("status", 1);
+        ret.put("message", "No matching service found.");
+        return ret;
+    }
+
+    private static JSONObject handleQuotaExhausted() {
+        JSONObject ret = new JSONObject(true);
+        ret.put("status", 3);
```
vibhcool: mentioning limit in config.properties would be better

src/org/loklak/api/search/VideoUrlService.java
singhpratyush: Compiling the regex every time the function is called would be not a good idea.

singhpratyush:

Achint08:
This link you have mentioned returns four videos.
-  The first link is a video with `.m3u8` extension. These links can be removed as we corresponding `.mp4` file extension videos. 

- Moreover, all the `.mp4` videos links provide the same video. Just the difference is that we have different dimensions for a single video.  Can you provide just one video with the maximum dimension?

singhpratyush:
@Achint08:

> The first link is a video with `.m3u8` extension. These links can be removed

`m3u8` is really great for streaming videos. Maybe we should keep it and would integrate it later in loklak search. Though this is low priority now, having it would provide better streaming experience.

> Can you provide just one video with the maximum dimension?

May not be desirable everytime. We should also consider users who have slower/limited internet connections.

Achint08:
@singhpratyush Great! 👍 :)
This seems considerable. Maybe in loklak search, we can provide an option for dimension.

vibhcool:
works fine, if video links are of standard 3 quality types and one m3u8, then maybe keys shall be there to mention types of links

singhpratyush:
@vibhcool: I don't think that it would be needed. To separate them with keys, we would simply be using link text and leaving it to clients would be more beneficial I guess.

vibhcool:
@singhpratyush how would user differentiate between m3u8 link and mp4 link? indulging in string-substring matching may not be in user's interest

singhpratyush:
@Achint08 @hemantjadon: From the user point of view, do you feel like having different format videos under different keys would be better?

hemantjadon:
@singhpratyush I guess untill the results are provided under one key, there is no need to specify types, our work is to provide the user with the links its his responsibility how to use that data

singhpratyush:
@hemantjadon: This PR is not deciding whether the endpoint will be mandatory to use for end users. It is just making the endpoint existent so that if in any case, anyone wants to use this service, they can do so.

Moreover, we can later improve this service to provide a redirect to actual video from this URL, just like the unshortening service.

singhpratyush:
@vibhcool: So @hemantjadon mentions that separate keys are not needed. Shall we skip it?

vibhcool:
IMHO, there should be different keys for different formats of content. 😅

vibhcool:
@hemantjadon , this endpoint is to fetch any video from any website. This is initially set up for twitter. This can be set up for other websites also

singhpratyush:
@vibhcool: But they are noth videos. Even the Twitter API returns similar response.

hemantjadon:
I agree with @singhpratyush  👍  Why make such distinctions I am unable to understand?

vibhcool:
@singhpratyush @hemantjadon , yep, both are videos but are processed differently. m3u8 isn't binary where as mp4 is binary.

singhpratyush:
@vibhcool @hemantjadon: Deciding this now could be tricky. I think we should introduce this change if anyone faces issues with "all video under one key" thing.

hemantjadon:
@singhpratyush I agree 👍

PR #1462, opened 2017-08-16
Remove hash from hashtag list

### LINKED ISSUE #1457: Hash symbol appended in front of hashtag set linked to a status and to the aggregations
- Issue type: Bug report

### Short description

The backend is returning all the entities with hash symbol appended, at the start of every string of hashtag set linked to a status and to the aggregations.
This behaviour is unexpected.

singhpratyush:
### Short description

Fixes #1457.

I have:

### For the reviewers
I have:

### DIFF
--- a/src/org/loklak/objects/MessageEntry.java
+++ b/src/org/loklak/objects/MessageEntry.java
@@ -57,7 +57,7 @@ public class MessageEntry extends AbstractObjectEntry {
     // left boundary must be space since the @ is itself a boundary
     public final static Pattern USER_PATTERN = Pattern.compile("(?:[ (]|^)(@..*?)(?:\\b|$)");
     // left boundary must be a space since the # is itself a boundary
-    final static Pattern HASHTAG_PATTERN = Pattern.compile("(?:[ (]|^)(#..*?)(?:\\b|$)");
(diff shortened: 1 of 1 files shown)


src/org/loklak/objects/MessageEntry.java
```diff
@@ -57,7 +57,7 @@
     // left boundary must be space since the @ is itself a boundary
     public final static Pattern USER_PATTERN = Pattern.compile("(?:[ (]|^)(@..*?)(?:\\b|$)");
     // left boundary must be a space since the # is itself a boundary
-    final static Pattern HASHTAG_PATTERN = Pattern.compile("(?:[ (]|^)(#..*?)(?:\\b|$)");
+    final static Pattern HASHTAG_PATTERN = Pattern.compile("(?:[ (]|^)#(..*?)(?:\\b|$)");
```
sudheesh001: Any tests? Doesn't keeping the `#` outside mandate that the hash is available in the pattern?

It'd always be better to have a test to pass and fail these patterns, that'd give a detailed reason as to why this change in regex is needed.

vibhcool: @sudheesh001 There isn't a bug in hashtags, the hashtags work fine, but Loklak Search team suggested that hashtags strings are more useful than hashtags with `#` symbol. https://github.com/loklak/loklak_server/issues/1457#issuecomment-322570756

Issue #1182, opened 2017-05-21
Prod Deploy: Move to SSL for https://api.loklak.org or loklak.org 


sudheesh001:
- Issue type: Feature request 

### Steps to reproduce

1. Move to support https://
2. HTTP 302 `http://` requests to `https://`

### Questions to consider
1. Should we serve the SSL Cert over Nginx?
2. Should we serve the SSL Cert while using Jetty as it exists?
 https://github.com/loklak/loklak_server/blob/development/conf/config.properties#L57

vibhcool:
According to conversation on Gitter, We can setup Nginx with Kubernetes to move to http with SSL. This will help us to get SSL certificates with auto renewal.

The Nginx will be at the ingress controller level and can be used for any project running on the same cluster. Like for example if we have a separate Loklak Frontend running, that will also go through the Nginx-based ingress controller.

The jetty will work at back as the server for Loklak Server.
Creating issue for this.

PR #1464, opened 2017-08-16
Fixes #1463: Set up Nginx with Kubernetes with SSL for master branch

### LINKED ISSUE #1463: Set up Nginx with Kubernetes with SSL for master branch
- Issue type: Feature request

This Issue is to adds SSL for master branch GKE deployment for api.loklak.org

vibhcool:
Fixes Issue #1463 , #1182 
Works according to Loklak Server logs and Nginx-Ingress logs
Output is default backend, http://35.189.158.143/ . Due to no access to host api.loklak.org, TLS certificate can't be generated. Rest works fine.

### Short description
I have:

### For the reviewers
I have:


kubernetes/bin/deploy.sh
```diff
@@ -1,53 +1,59 @@
...
-USAGE="USAGE: ./kubernetes/bin/deploy-development.sh create|create all|delete|delete all"
+USAGE="USAGE: ./kubernetes/bin/deploy.sh  api|staging create|create all|delete|delete all"
 
-if [[ "$1" = "create" ]];
+if [[ "$1" = "api" ]]
 then
+    export BRANCH=api
+else
+    export BRANCH=staging
+fi
```
singhpratyush: Why not simply
```bash
BRANCH=$1
```

Why are you exporting this?

kubernetes/bin/deploy.sh
```diff
@@ -1,53 +1,59 @@
...
-    kubectl create -R -f ${DIR}/yamls/staging/web/
+    kubectl create -R -f ${DIR}/yamls/${BRANCH}/web/
     # Wait for some time to prevent any chance of API-Server to access incomplete deployments
     sleep 20
-    echo "Creating loklak deployment"
+    echo "Creating Loklak Server deployment"
     # Create API server deployment and service for development branch
-    kubectl create -R -f ${DIR}/yamls/staging/api-server/
-    echo "Deployed loklak on Kubernetes"
+    kubectl create -R -f ${DIR}/yamls/${BRANCH}/api-server/
```
singhpratyush: We would need to set some different name/namespace for staging and api deployments as they would conflict while updating images via Travis.

kubernetes/yamls/api/api-server/api-service.yml
```diff
@@ -11,3 +11,4 @@ spec:
     targetPort: 80
   selector:
     app: server
+  type: LoadBalancer
```
niranjan94: Not needed since it will go via the Nginx IC

kubernetes/bin/deploy.sh
vibhcool: So that if we want we can change the name of directories if needed.
If `$1` is set as any other random directory name, then setup will just fail with inappropriate errors. I think else condition should be to print error. I will fix this.

kubernetes/bin/deploy.sh
vibhcool: I didn't understand 😅

niranjan94: yes. Should be in different namespaces.

vibhcool: fixed this :)

kubernetes/bin/deploy.sh
```diff
@@ -1,53 +1,64 @@
 #!/bin/bash
 export DIR=kubernetes
 
-USAGE="USAGE: ./kubernetes/bin/deploy-development.sh create|create all|delete|delete all"
+USAGE="USAGE: ./kubernetes/bin/deploy.sh  api|staging create|create all|delete|delete all"
 
-if [[ "$1" = "create" ]];
+if [[ "$1" = "api" ]]
 then
+    export BRANCH=api
```
singhpratyush: I still don't get the intention while exporting this, and we can simply set
```bash
BRANCH=$1
```

kubernetes/bin/update-image.sh
```diff
@@ -0,0 +1,19 @@
...
+then
+    BRANCH=Staging
+    IMAGE=loklak/loklak_server:latest-kubernetes-staging
+else
+    IMAGE=$1
+fi
+
+echo "Setting image $IMAGE"
+
+kubectl set image deployment/server --namespace=api-web server=$IMAGE
```
singhpratyush: Why `api-web`?

PR #1507, opened 2017-08-26
Increase search timeout in config

### LINKED ISSUE #1506: Increase timeout for search
### Short description

The search endpoint works based on the timeout specified in config.properties. In order to perform more complex searches, it is needed to increase timeout so that the results are given time to complete.

singhpratyush:
### Short description

Fixes #1506.

I have:

### For the reviewers
I have:

### DIFF
--- a/conf/config.properties
+++ b/conf/config.properties
@@ -93,7 +93,7 @@ search.count.low = 10
 search.count.default = 20
 search.count.max.public = 100
 search.count.max.localhost = 1000
-search.timeout = 3000
+search.timeout = 5000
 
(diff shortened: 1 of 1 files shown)


singhpratyush:

mariobehling:
Why are these things still hardcoded? What options are there to configure this, e.g. in environment variables? How to avoid hardcoding things. We should be moving away from this as this should be configured on the server.

singhpratyush:
@mariobehling: This option is kept in the configuration file which can be changed by users at their will. It is similar to making the settings as environment variables, just that these settings are stored in a file here.

Issue #1353, opened 2017-07-23
API endpoint to check if a tweet is collected


yukiisbored:
- Issue type: Feature request 

An API endpoint that checks if a tweet is collected is helpful for those who want a tweet to be scanned and archived into Loklak's database in the same way on how the "Wayback Machine" works.

The API can look something familiar like below:
  * `/api/show.json?tweet=https://twitter.com/ArtisticBondage/status/885041783079661569`
  * `/api/show.json?tweet=885041783079661569`
  * `/api/show.json?user=https://twitter.com/ArtisticBondage`
  * `/api/show.json?user=ArtisticBondage`

The API does the following:
  * If it's a tweet, check if it exists or not
    * If it exists, show it.
    * If not, Scrap the tweet and add it to the db, return the same thing if it exists
  * If it's a user, check if we have anything by the user or not
    * If it exists, check how many tweets are archived from that user and archive the tweets that aren't archived yet and return what we know about the user and possibly the tweets (the follower count, etc)
    * If it doesn't exist, archive the tweets and the info about the user, return the same thing if it exists.

Possible use cases:
  * Archiving
  * Passive-way to receive new tweets

vibhcool:
- Archiving is nice idea. It can be implemented by improving tweets dump implementation.

- Run the api endpoint, if the tweet related to info exists, exit, else collect the tweet?

AnshulMalik:
I'd like to give it a try :)

yukiisbored:
@AnshulMalik I've invited you to this org so I can assign you to this issue. Can you approve the invitation?

Thanks

yukiisbored:
@vibhcool Yes, that's the idea. We can make Loklak into an archiving utility for tweets.

AnshulMalik:
@yukiisbored I am in :)

I'll soon create a PR of my first attempt, then we can iterate over to build the feature.

vibhcool:
@yukiisbored , but this feature exists in Loklak Server. dumps can be downloaded from `/dump.html` 😅

yukiisbored:
@vibhcool But I can't see what those dumps contain and I can't see if they actually have the tweet I want nor I can request to archive a tweet ASAP.

PR #1527, opened 2017-09-11
[WIP] Check whether a tweet is scraped

AnshulMalik:
Issue: #1353 

### Short description
Initial steps to add the new feature
- Created a new servlet to handle new endpoint `check.json`


vibhcool:
@AnshulMalik, what is the possible usecase for checking if tweet is scraped? It can be checked on quering from cache. there is a existing method to check if tweet exist. 
the issue is to scrape tweets and output as archive from search endpoint.

vibhcool:
you may discuss in issue or gitter, how to approach for the solution before implementing it :)

src/org/loklak/LoklakServer.java
```diff
@@ -104,25 +104,7 @@
...
-import org.loklak.api.search.ShortlinkFromTweetServlet;
-import org.loklak.api.search.SuggestServlet;
-import org.loklak.api.search.TimeAndDateService;
-import org.loklak.api.search.TweetScraper;
-import org.loklak.api.search.UserServlet;
-import org.loklak.api.search.VideoUrlService;
-import org.loklak.api.search.WeiboUserInfo;
-import org.loklak.api.search.WikiGeoData;
-import org.loklak.api.search.WordpressCrawlerService;
+import org.loklak.api.search.*;
```
sudheesh001: No! We didn't do this for a reason. Please undo this change.

src/org/loklak/api/search/TweetCheckServlet.java
```diff
@@ -0,0 +1,66 @@
...
+        String tweetQuery = query.get("tweet", "");
+        String userQuery = query.get("user", "");
+
+        response.setCharacterEncoding("UTF-8");
+
+        PrintWriter sos = response.getWriter();
+
+        JSONObject result = new JSONObject(true);
+
+        if(tweetQuery != null) {
```
sudheesh001: space consistency. `if (` instead of `if(`

src/org/loklak/api/search/TweetCheckServlet.java
```diff
@@ -0,0 +1,66 @@
...
+            String tweetId = parseTweetId(tweetQuery);
+            boolean exists = DAO.existMessage(tweetId);
+            if(exists) {
```
sudheesh001: Same space consistency here.

src/org/loklak/api/search/TweetCheckServlet.java
```diff
@@ -0,0 +1,66 @@
...
+            }
+        }
+
+        query.setResponse(response, jsonp ? "application/javascript" : "application/json");
+
+        if (jsonp) sos.print(callback + "(");
+        sos.print(result.toString(minified ? 0 : 2));
+        if (jsonp) sos.println(");");
+        sos.println();
+        sos.flush(); sos.close();
```
sudheesh001: `sos.close()` on new line.

src/org/loklak/api/search/TweetCheckServlet.java
```diff
@@ -0,0 +1,66 @@
...
+        Query query = RemoteAccess.evaluate(request);
+
+        // manage DoS
+        if (query.isDoS_blackout()) {response.sendError(503, "your request frequency is too high"); return;}
+
+        String callback = query.get("callback", "");
+        boolean jsonp = callback != null && callback.length() > 0;
+        boolean minified = query.get("minified", false);
+
+        String tweetQuery = query.get("tweet", "");
```
sudheesh001: Does the user put the entire link here? or just the `tweetid`? If it's the entire link, why is it designed this way? In what scenarios will someone be using this API call with a `full` URL string.

In ideal cases if you were looking up something, you'd look things up by ID and not by the entire path to the ID right, the same should be the analogy here. Making this just the `tweetID:string` will make `parseTweetId()` a redundant method.

sudheesh001:
A little more thought experiment
- [ ] What should the API do when it doesn't find the required string? (Add link to the queue to crawl/scrape; Validate that a tweet exists on twitter but we don't have it indexed yet?)
- [ ] What is the rationale behind using the entire URL in the parameters to be passed? As pointed out in the reviews above.

src/org/loklak/LoklakServer.java
AnshulMalik: Ok but May I know the reason?

src/org/loklak/api/search/TweetCheckServlet.java
AnshulMalik: The issue said: 
- `/api/show.json?tweet=https://twitter.com/ArtisticBondage/status/885041783079661569`
- `/api/show.json?tweet=885041783079661569`

Now if it's a link, we extract the id, and if it's already id, we are good to go.

So I added this method

src/org/loklak/LoklakServer.java
sudheesh001: The main reason why we do not do this is future compatibility, if there's a package which says `a.b.c.*` and `x.y.z.*` both having an class of say `Person` then there are higher chances for errors and the compilers getting confused. Also using the `*` means you're increasing the lookup time for the compiler while compilation by a small time but the resulting byte code will remain the same anyway.

I agree that having a long list of the imports can be quite intimidating but good IDEs always collapse them and you do not have to worry about what you import either because if there's a class you're using which isn't found in the file as an import, the IDE allows you to press a shortcut and it'll auto import that for you.

AnshulMalik: Thanks
Tip: This change was done be intellij itself :)

src/org/loklak/api/search/TweetCheckServlet.java
sudheesh001: The question here would be how would you know from the current method that the `tweet` value is a valid ID or a link. Doing a check for `http` contained in the string would not help here at all. I can pass my string to be `tweet=https://126312356123612/ajsdba/1 which will pass and it'll treat the ID as `1`  but this has major problems because the URL isn't by itself a valid URL.

You need to parse `https://twitter.com/{:user}/status/{:id}` for validation.

sudheesh001:
left a comment at https://github.com/loklak/loklak_server/pull/1527#discussion_r138239339 The PR still doesnot adequately address the issue.
This questions isn't answered too
- [ ] What should the API do when it doesn't find the required string? (Add link to the queue to crawl/scrape; Validate that a tweet exists on twitter but we don't have it indexed yet?)

mariobehling:
@AnshulMalik Any update on this? Would be great to see this merged.

@sudheesh001 If there is no other activity should this be closed?

PR #1551, opened 2017-10-21
merge development into master

Orbiter:
### Short description
I have:
- [ ] There is a corresponding issue for this pull request.
- [ ] Mentioned the Issue number in the pull request commit message `Fixes #<number> commit message`
- [ ] There is only strictly only one commit per issue.

### For the reviewers
I have:
- [ ] Reviewed this pull request by an authorized contributor.
- [ ] The reviewer is assigned to the pull request.

PR #1554, opened 2017-10-28
Loklak move repeated version to gradle properties

mrnakumar:
Currently there are two things which are not so good about build.gradle. 

1. Same version number is repeated multiple times. For example we have dependency on lucene-memory and lucene-query, even though we are using same version of both of these. we repeat the version number with each of these two entries in build.gradle, someting like show below:-
"org.apache.lucene:lucene-memory:5.5.2"
"org.apache.lucene:lucene-queries:5.5.2"

here version number "5.5.2" repeats, which is not good. This pull request make it like this:
"org.apache.lucene:lucene-memory:${luceneVersion}"
"org.apache.lucene:lucene-queries:${luceneVersion}"

where luceneVersion is provided in gradle.properties

2. We are using verbose form of dependency which is difficult to read. 
it is always better to use : 'group:artifact:version' compared to 'group : g1,artifact:a1,version:v1'


singhpratyush:
The build is failing due to the following exception. Please take a look.

```
FAILURE: Build failed with an exception.
* Where:
Build file '/loklak_server/build.gradle' line: 62
* What went wrong:
A problem occurred evaluating root project 'loklak_server'.
> Could not get unknown property 'luceneVersion' for object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.
```

sudheesh001:
Hey @mrnakumar Thank you for your first contribution. However, there seem to be a few issues and the build needs to pass before we take further steps to merge the pull request. At the same time, we encourage you to have made just 1 commit to fix a specific issue. Currently there seems to be 4 of them, one good way to overcome that is to rebase and squash the commits into one single commit.

Feel free to reach out to the chat on gitter and the community members will be able to definitely help you out.

mrnakumar:
Thanks @sudheesh001 . I have squashed the commits into one. Now you should see only one commit. 

Regarding the build failure: I  see that build is failing because the 'git' is not installed on the docker image but the file 'bin/start.sh' needs it, I have pasted the two lines from the file below:-
# Set commit details as env var
export COMMIT_HASH=`git rev-parse HEAD`
export COMMIT_COMMENT=`git log -1 --pretty=%B`

I was wondering, if this thing is really required? I see one more pull request is failing because of the same reason(merge development into master #1551).  @singhpratyush Any comment on this?

PR #1642, opened 2020-07-26
Checks for user key using json has instead of an explicit read and check

apatruni:
### Short description
I have:
- [ ] There is a corresponding issue for this pull request.
- [ ] Mentioned the Issue number in the pull request commit message `Fixes #<number> commit message`
- [ ] There is only strictly only one commit per issue.

### For the reviewers
I have:


sudheesh001 (changes requested):
This looks good. However I'd suggest a few changes before this can be merged.

Would it be possible to change the commit message to:
`Checks for user key using json has instead of an explicit read and check`

With the change in commit message, I think this can be approved. 

Handing over the second review to @Orbiter

revmarcuspeter-cyber (approved):

revmarcuspeter-cyber (approved):

